PR #4, opened 2023-01-24
Store name of testsetup in testitem

nickrobinson251:


nickrobinson251:
before this would error with
```
julia> @testitem "Foo" setup=[FooSetup] begin
           @test true
       end
ERROR: UndefVarError: FooSetup not defined
Stacktrace:
 [1] top-level scope
   @ ~/repos/ReTestItems.jl/src/ReTestItems.jl:94
```

quinnj (commented):
Thanks!

PR #9, opened 2023-01-25
Add TestEnv as dep and do activation when running test items

quinnj:


src/ReTestItems.jl
```diff
@@ -194,47 +211,71 @@ function runtests end
...
-    fch = FilteredChannel(shouldrun, chan(ch))
-    @debug "Including tests in $dirs."
-    errormonitor(@spawn include_testfiles!($(setupctx.setups), $fch, dirs...))
-    @debug "Scheduling tests on $(nprocs()) processes."
-    if nprocs() == 1
-        schedule_testitems!(setupctx, chan(ch))
-    else
-        futures = [@spawnat(p, schedule_testitems!(setupctx, ch)) for p in workers()]
-        foreach(wait, futures)
+    # identify project directory for which we're running tests
```
nickrobinson251, line 237: maybe pull this out into a function so we can add a couple unit tests?

src/ReTestItems.jl
```diff
@@ -194,47 +211,71 @@ function runtests end
...
+    projectfile = ""
+    for dir in dirs
+        path = dir
```
nickrobinson251, line 240: probably need an `abspath` somewhere, else we could hit `dirname("")` forever... maybe here?
```suggestion
        path = abspath(dir)
```

test/packages/NoDeps.jl/Project.toml
```diff
@@ -2,6 +2,10 @@ name = "NoDeps"
 uuid = "2520ce14-60c1-53a9-8c5b-16f535851c44"
 version = "0.0.0"
 
+[deps]
+ReTestItems = "817f1d60-ba6b-4fd5-9520-3cf149f6a823"
+TestEnv = "1e6cf692-eddd-4d53-88a5-2d735e33781b"
```
nickrobinson251, line 7: this shouldn't need to depend on these, i don't think

nickrobinson251 (approved):
looks good 👍

PR #15, opened 2023-03-23
Add compile times to collected stats

Drvi:
image


Drvi:
1.6 CI fails because of the use of `@atomic`, nighly fails because test Manifests were created with julia 1.8 and this emits warn messages where we don't expect them to be.

PR #24, opened 2023-04-01
Test we support Julia v1.9 

nickrobinson251:
- #22

### DIFF
--- a/test/integrationtests.jl
+++ b/test/integrationtests.jl
@@ -54,34 +54,26 @@ end
 
     # warn if the path does not exist
     dne = joinpath(pkg, "does_not_exist")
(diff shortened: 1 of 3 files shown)


test/integrationtests.jl
```diff
@@ -54,34 +54,26 @@ end
 
     # warn if the path does not exist
     dne = joinpath(pkg, "does_not_exist")
-    @test_logs (:warn, "No such path \"$dne\"") begin
+    @test_logs (:warn, "No such path \"$dne\"") match_mode=:any begin
```
quinnj, line 57: Oh man, this is so much better; didn't know about `match_mode`!

quinnj (approved):

PR #25, opened 2023-04-03
Add memory consumption and reformat stats

Drvi:
* removes some stale logging code from the multithreaded days
* adds memory logs to non-interactive sessions
* makes the stats a bit more succinct

### DIFF
--- a/src/log_capture.jl
+++ b/src/log_capture.jl
@@ -53,15 +53,15 @@ function time_print(io; elapsedtime, bytes=0, gctime=0, allocs=0, compile_time=0
     if  gctime > 0 || compile_time > 0
         print(io, " (")
         if compile_time > 0
-            _print_scaled_one_dec(io, 100compile_time, elapsedtime, "% compile")
+            _print_scaled_one_dec(io, 100 * compile_time, elapsedtime, "% compile")
         end
         if recompile_time > 0
             print(io, ", ")
-            _print_scaled_one_dec(io, 100recompile_time, elapsedtime, "% recompile")
+            _print_scaled_one_dec(io, 100 * recompile_time, elapsedtime, "% recompile")
         end
         if gctime > 0
             compile_time > 0 && print(io, ", ")
-            _print_scaled_one_dec(io, 100gctime, elapsedtime, "% GC")
+            _print_scaled_one_dec(io, 100 * gctime, elapsedtime, "% GC")
         end
         print(io, ")")
     end
@@ -114,9 +114,12 @@ _has_logs(ts::TestSetup) = filesize(logpath(ts)) > 0
 # The path might not exist if a testsetup always throws an error and we don't get to actually
 # evaluate the test item.
 _has_logs(ti::TestItem, i=nothing) = (path = logpath(ti, i); (isfile(path) && filesize(path) > 0))
+# Stats to help diagnose OOM issues.
 _mem_watermark() = string(
-    "maxrss ", lpad(Base.Ryu.writefixed(100Float64(Sys.maxrss()/Sys.total_memory()), 1), 4),
-    "% | mem ", lpad(Base.Ryu.writefixed(100Float64(1 - (Sys.free_memory()/Sys.total_memory())), 1), 4),
+    # Tracks the peak memory usage of a process / worker
+    "maxrss ", lpad(Base.Ryu.writefixed(100 * Float64(Sys.maxrss()/Sys.total_memory()), 1), 4),
+    # Total memory pressure on the machine
+    "% | mem ", lpad(Base.Ryu.writefixed(100 * Float64(1 - (Sys.free_memory()/Sys.total_memory())), 1), 4),
     "% | "
 )
 
--- a/src/log_capture.jl
+++ b/src/log_capture.jl
@@ -25,38 +25,56 @@ function default_log_display_mode(report::Bool, nworkers::Integer, interactive::
     end
 end
 
-# Adapted from Base.time_print
+# Adapted from timing.jl
+const _cnt_units = ["", " K", " M", " B", " T", " P"]
+const _mem_pow10_units = ["byte", "KB", "MB", "GB", "TB", "PB"]
+function _format_pow10_bytes(bytes)
+    bytes, mb = Base.prettyprint_getunits(bytes, length(_mem_pow10_units), Int64(1000))
+    if mb == 1
+        return string(Int(bytes), " ", _mem_pow10_units[mb], bytes==1 ? "" : "s")
+    else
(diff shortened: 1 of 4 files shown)


src/log_capture.jl
```diff
@@ -86,18 +104,6 @@ function _redirect_logs(f, target::IO)
     redirect_stdio(f, stdout=colored_io, stderr=colored_io)
 end
 
-# A lock that helps to stagger prints to DEFAULT_STDOUT, e.g. when there are log messages
-# comming from distributed workers, reports for stalled tests and outputs of
-# `print_errors_and_captured_logs`.
-const LogCaptureLock = ReentrantLock()
-macro loglock(expr)
-    return :(@lock LogCaptureLock $(esc(expr)))
```
nickrobinson251, line 94: Why don't we need this anymore? I thought we would, because on the coordinator we `@spawn start_and_manage_worker` for each worker, so even though we aren't running each testitem on their own process, the coordinator could still have multiple threads processing results / printing logs (but maybe i'm misunderstanding)

src/log_capture.jl
```diff
@@ -108,6 +114,11 @@ _has_logs(ts::TestSetup) = filesize(logpath(ts)) > 0
 # The path might not exist if a testsetup always throws an error and we don't get to actually
 # evaluate the test item.
 _has_logs(ti::TestItem, i=nothing) = (path = logpath(ti, i); (isfile(path) && filesize(path) > 0))
+_mem_watermark() = string(
+    "maxrss ", lpad(Base.Ryu.writefixed(100Float64(Sys.maxrss()/Sys.total_memory()), 1), 4),
+    "% | mem ", lpad(Base.Ryu.writefixed(100Float64(1 - (Sys.free_memory()/Sys.total_memory())), 1), 4),
+    "% | "
```
nickrobinson251, line 120: i wonder if we could print this even more succinctly

it's nice that it's on the left of the RUNNING/FINISHED statement because (i) it keeps things fixed width, so logs line up nicely and (ii) it is more a property of the machine at that time that the specific testitem

but, it does mean the `item "foo"` statement is way to the right of the screen
```
Worker 2983:  02:39:53 | maxrss  7.4% | mem 16.4% | RUNNING  (  1/154) item "foo"
Worker 2983:  02:40:48 | maxrss  7.4% | mem 21.4% | FINISHED (  1/154) item "foo"
```

I wonder if we at least put them in the same column?

```
Worker 2983:  02:39:53 | maxrss  7.4% mem 16.4% | RUNNING  (  1/154) item "foo"
Worker 2983:  02:40:48 | maxrss  7.4% mem 21.4% | FINISHED (  1/154) item "foo"
```

src/log_capture.jl
```diff
@@ -108,6 +114,11 @@ _has_logs(ts::TestSetup) = filesize(logpath(ts)) > 0
 # The path might not exist if a testsetup always throws an error and we don't get to actually
 # evaluate the test item.
 _has_logs(ti::TestItem, i=nothing) = (path = logpath(ti, i); (isfile(path) && filesize(path) > 0))
+_mem_watermark() = string(
+    "maxrss ", lpad(Base.Ryu.writefixed(100Float64(Sys.maxrss()/Sys.total_memory()), 1), 4),
```
nickrobinson251, line 118:
```suggestion
    "maxrss ", lpad(Base.Ryu.writefixed(100 * Float64(Sys.maxrss()/Sys.total_memory()), 1), 4),
```

src/log_capture.jl
```diff
@@ -108,6 +114,11 @@ _has_logs(ts::TestSetup) = filesize(logpath(ts)) > 0
...
+    "% | mem ", lpad(Base.Ryu.writefixed(100Float64(1 - (Sys.free_memory()/Sys.total_memory())), 1), 4),
```
nickrobinson251, line 119:
```suggestion
    "% | mem ", lpad(Base.Ryu.writefixed(100 * Float64(1 - (Sys.free_memory()/Sys.total_memory())), 1), 4),
```

src/log_capture.jl
```diff
@@ -108,6 +114,11 @@ _has_logs(ts::TestSetup) = filesize(logpath(ts)) > 0
 # The path might not exist if a testsetup always throws an error and we don't get to actually
 # evaluate the test item.
 _has_logs(ti::TestItem, i=nothing) = (path = logpath(ti, i); (isfile(path) && filesize(path) > 0))
+_mem_watermark() = string(
```
nickrobinson251, line 117: Please can you add some docs, maybe just as a comment on this function, for what `maxrss` is as i expect lots of people not to know what it is / why it's useful info (at least i didn't 😛)

nickrobinson251 (approved):
Memory printing LGTM

Question about the log lock

src/log_capture.jl
Drvi: The lock was only used it in three places: `log_running`, `log_finished` and `print_errors_and_captured_logs`. The first two happened sequentially on each worker and the last one sequentially on the coordinator. So the way this was setup was not really helping anything. We might want to use a mechanism like this for the spawned tasks reading messages from the workers, as you mentioned, but since we are currently reading the output from the worker process line by line a more sophisticated strategy would be needed (like reading all available buffered bytes from the process, not just a line, making sure we handle the case where the buffer is full).

I think let's revisit this if we notice problems in practice.

src/log_capture.jl
nickrobinson251: sounds good to me

src/log_capture.jl
Drvi: Ugh, you are right... I forgot this changed with the worker PR. Let me try playing with it with :batched logs :+1:.

Drvi: I think it's on the safer side to keep it in... thanks for noticing!

quinnj:
Awesome stuff @Drvi. I would like to get your take on these new stats on some raicode runs from recent master. The [unit tests](https://hydra.dev.relational.ai/log/xqgnq7izq973jjssvy58kfmf1bbpgw7m-raicode-tests.unit.drv) seem to use a much higher rss/% total while running, which isn't surprising because we're using such a small machine. But it's curious that the [heavy tests](https://hydra.dev.relational.ai/log/skwnw3zc9kpf65y7kywqr4b2cb6l8w5x-raicode-tests.heavy.drv) percentages are so much lower. It makes me wonder if we could possibly go to a smaller machine and be ok, but I'm scared to because all my previous experiments ran into so many problems going smaller.

Drvi:
@quinnj I'd be interested to see a run, where the worker is OOM killed, just to see how useful this is. But yeah it seems to me, that the max RSS is not that high -- maybe `r6i.8xlarge` with 16 workers would be a good experiment?

quinnj:
Yeah, I agree it's worth trying

PR #27, opened 2023-04-03
Search for tests in all directories, not just `src` and `test`

nickrobinson251:


quinnj:
Looks like the `@testsetup` always included tests are failing though.

PR #33, opened 2023-04-05
Always start new worker after timeout or crash

nickrobinson251:
https://relationalai.atlassian.net/browse/RAI-11670


quinnj (approved):
Oh shoot, good catch. Darn these things are hard to catch in reviewing. Yeah, it seems like even just adding 1 more test to run after the existing abort or timeout testsets would be enough to ensure this works because we're running those with 1 worker, so we need to make sure any tests after the crashing one still get run.

PR #39, opened 2023-04-12
fix bounds error

rntz:
Fix an out-of-bounds error that could be caused, eg. by scheduling 8 testitems across 5 workers. The issue was in `get_starting_testitems`, which didn't check whether the indices it generated were in-bounds. Also fixes an issue where (eg) scheduling 8 items across 5 workers only actually uses 4 workers! The previous indexing arithmetic picked only 4 starting items: [1,3,5,7], then padded with `nothing` up to 5 items. The worker that gets `nothing` then immediately quits.


rntz:
this seems to be failing a CI test but I don't understand why

quinnj (approved):
LGTM; don't worry about the test failure, I think it's just a flaky/timing-dependent test.

nickrobinson251:
Can we have tests please?

rntz:
I'd love to add tests but I don't really understand your testing setup here.

nickrobinson251:
I've made a PR to document the testing setup: https://github.com/JuliaTesting/ReTestItems.jl/pull/43

For a change like this there are a couple options:
- probably best, if easy: write a unit test for `get_starting_testitems` by manually construct a `TestItems` and setting a `nworkers` that triggers the issue you were seeing and test that the code now has the expected behaviour e.g.
    ```julia
    graph = ReTestItems.FileNode("")  # we don't use the graph info for this test
    # we previously saw `BoundsError` with 8 testitems, 5 workers
    testitems = [@testitem "ti-$i" begin; end; for i in 1:8]
    nworkers = 5
    starts = get_starting_testitems(TestItems(graph, testitems, 0), nworkers)
    @test length(starts) == 8
    @test allunique(map(ti -> ti.name, starts))
    ```
    - i'd probably put this in `internals.jl`
- and/or write an "integration test" for `runtest` by creating a file with a number of testitems, named something like `_starting_testitems_test.jl` and then in `integration_tests.jl` call `runtests("_starting_testitems_test.jl"; nworkers)` and test you get the expected results (using `encased_testset` to capture the results of `runtests`), e.g.
    ```julia
    results = encased_testset() do
        runtests(joinpath(_TEST_DIR, "_starting_testitems_test.jl"); nworkers=5)
    end
    # `_starting_testitems_test.jl` has 8 test items
    # we previously saw `BoundsError` in `get_starting_testitems` with 8 testitems, 5 workers
    @test n_passed(results) == 8
```

rntz:
Thanks @nickrobinson251 . I've added unit tests for `get_starting_tetitems` for 1-10 workers across 1-10 items.

Issue #32, opened 2023-04-05
Add tests for logger world age fix


nickrobinson251:
we wrap our `@spawn` tasks in 
```julia
with_logger(current_logger()) do
   ...
```
so we use the logger that was set before we `eval`'d any user code (which may define a custom logger), to avoid world age issues when logging https://github.com/JuliaLang/julia/issues/33865

But i don't think we've any tests to make sure we don't re-introduce this issue. We should add a test.

---

The issues look like, e.g.

```julia
ERROR: MethodError: no method matching shouldlog(::CustomLogging.LocalLogger, ::Base.CoreLogging.LogLevel, ::Module, ::Symbol, ::Symbol)
The applicable method may be too new: running in world age 32437, while current world is 33036.
Closest candidates are:
  shouldlog(::CustomLogging.LocalLogger, ::Any, ::Any, ::Any, ::Any) at ~/repos/CustomLogging/src/local_logger.jl:52 (method too new to be called from this world context.)
  shouldlog(::LoggingExtras.ActiveFilteredLogger, ::Any...) at ~/julia_depots/LoggingExtras/zT9ZU/src/activefiltered.jl:37
  shouldlog(::LoggingExtras.MinLevelLogger, ::Any...) at ~/julia_depots/LoggingExtras/zT9ZU/src/minlevelfiltered.jl:21
  ...
Stacktrace:
  [1] _invoked_shouldlog
    @ ./logging.jl:78 [inlined]
  [2] macro expansion
    @ ./logging.jl:381 [inlined]
  [3] _runtests_in_current_env(shouldrun::Function, paths::Tuple{String}, projectfile::String, nworkers::Int64, nworker_threads::Int64, testitem_timeout::Int64, verbose::Bool, debug::Int64)
    @ ReTestItems 
    ...
```

PR #46, opened 2023-05-02
Wrap `_runtests_in_current_env` in current logger to avoid world age issues

nickrobinson251:
fixes the first issue in https://relationalai.atlassian.net/browse/RAI-12275  -- tested to work locally

(related to #32 but doesn't actually add a test because it is surprisingly difficult to test, and also this issue is fixed upstream on master https://github.com/JuliaLang/julia/issues/33865#issuecomment-1527900246)


quinnj (approved):

NHDaly:
Thanks both! :)

PR #52, opened 2023-05-12
Fix the GC.gc() invocation to do a full mark+sweep

NHDaly:
The `true` param tells the GC to do a full _sweep_, which clears out all the mark bits, and then the _next_ GC will be a full mark, since all the bits have been cleared.


NHDaly:
oh huh i guess i never opened this PR. whoops :)

quinnj (approved):

PR #54, opened 2023-05-15
Support creating a `JUnitTestCase` from a `DefaultTestSet` alone

nickrobinson251:
- in particular make `PerfStats` optional
- cc @mcmcgrath13

### DIFF
--- a/src/junit_xml.jl
+++ b/src/junit_xml.jl
@@ -28,7 +28,7 @@ end
 mutable struct JUnitTestCase  # TestItem run
     const name::String
     counts::JUnitCounts
-    stats::PerfStats
+    stats::Union{PerfStats, Nothing} # Additional stats not available from the testset
     error_message::Union{String,Nothing} # Additional message to include in `<error>`
     logs::Union{Vector{UInt8},Nothing} # Captured logs from a `@testitem`
 end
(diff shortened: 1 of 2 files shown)


nickrobinson251:
xref https://relationalai.atlassian.net/browse/RAI-12694

quinnj (approved):

src/junit_xml.jl
```diff
@@ -28,7 +28,7 @@ end
 mutable struct JUnitTestCase  # TestItem run
     const name::String
     counts::JUnitCounts
-    stats::PerfStats
+    stats::Union{PerfStats, Nothing} # Additional stats not available from the testset
     error_message::Union{String,Nothing} # Additional message to include in `<error>`
     logs::Union{Vector{UInt8},Nothing} # Captured logs from a `@testitem`
```
mcmcgrath13, line 33: do logs need to be a vector of `UInt8`? Or is a string okay here? The way my logs are captured they are of type `String`

mcmcgrath13 (commented):
Thank you!

src/junit_xml.jl
mcmcgrath13: Actually, I take that back. I end up writing everything to a buffer when it gets formatted, so it is a `Vector{UInt8}`

PR #56, opened 2023-05-18
Save and reuse when possible the temp test environments

quinnj:
This avoids unnecessary recompilation/reactivation of test environments when a user repeatedly calls `runtests` on the same project from the REPL. The temp environments are created via `mktempdir` in TestEnv.jl, so they'll auto be cleaned up for us when the process exits.

### DIFF
--- a/src/ReTestItems.jl
+++ b/src/ReTestItems.jl
@@ -224,7 +224,6 @@ function _runtests(shouldrun, paths, nworkers::Int, nworker_threads::Int, worker
                     testenv = TestEnv.activate()
                     TEST_ENVS[proj_file] = testenv
                 end
-                @debugv 2 "Running in test environment `$(Base.active_project())`"
                 _runtests_in_current_env(shouldrun, paths, proj_file, nworkers, nworker_threads, worker_init_expr, testitem_timeout, retries, verbose_results, debug, report, logs)
             finally
                 Base.set_active_project(orig_proj)
--- a/src/ReTestItems.jl
+++ b/src/ReTestItems.jl
@@ -214,14 +214,26 @@ function _runtests(shouldrun, paths, nworkers::Int, nworker_threads::Int, worker
             return _runtests_in_current_env(shouldrun, paths, proj_file, nworkers, nworker_threads, worker_init_expr, testitem_timeout, retries, verbose_results, debug, report, logs)
         else
             @debugv 1 "Activating test environment for `$proj_file`"
-            return Pkg.activate(proj_file) do
-                TestEnv.activate() do
-                    _runtests_in_current_env(shouldrun, paths, proj_file, nworkers, nworker_threads, worker_init_expr, testitem_timeout, retries, verbose_results, debug, report, logs)
+            orig_proj = Base.active_project()
+            try
+                if haskey(TEST_ENVS, proj_file)
+                    testenv = TEST_ENVS[proj_file]
+                    Base.set_active_project(testenv)
+                else
+                    Pkg.activate(proj_file)
+                    testenv = TestEnv.activate()
+                    TEST_ENVS[proj_file] = testenv
                 end
+                @debugv 2 "Running in test environment `$(Base.active_project())`"
+                _runtests_in_current_env(shouldrun, paths, proj_file, nworkers, nworker_threads, worker_init_expr, testitem_timeout, retries, verbose_results, debug, report, logs)
+            finally
+                Base.set_active_project(orig_proj)
             end
         end
     end
 end
+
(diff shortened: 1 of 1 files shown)


src/ReTestItems.jl
```diff
@@ -214,15 +214,27 @@ function _runtests(shouldrun, paths, nworkers::Int, nworker_threads::Int, worker
...
+            try
+                if haskey(TEST_ENVS, proj_file)
+                    testenv = TEST_ENVS[proj_file]
+                    Base.set_active_project(testenv)
+                else
+                    Pkg.activate(proj_file)
+                    testenv = TestEnv.activate()
+                    TEST_ENVS[proj_file] = testenv
                 end
+                @debugv 2 "Running in test environment `$(Base.active_project())`"
```
nickrobinson251, line 227:
```suggestion
```

src/ReTestItems.jl
```diff
@@ -214,15 +214,27 @@ function _runtests(shouldrun, paths, nworkers::Int, nworker_threads::Int, worker
...
+                _runtests_in_current_env(shouldrun, paths, proj_file, nworkers, nworker_threads, worker_init_expr, testitem_timeout, retries, verbose_results, debug, report, logs)
+            finally
+                Base.set_active_project(orig_proj)
             end
         end
     end
 end
 
+const TEST_ENVS = Dict{String, String}()
```
nickrobinson251, line 236: style-wise, i'd probably put this above the function using it

src/ReTestItems.jl
```diff
@@ -214,15 +214,27 @@ function _runtests(shouldrun, paths, nworkers::Int, nworker_threads::Int, worker
             return _runtests_in_current_env(shouldrun, paths, proj_file, nworkers, nworker_threads, worker_init_expr, testitem_timeout, retries, verbose_results, debug, report, logs)
         else
             @debugv 1 "Activating test environment for `$proj_file`"
-            return Pkg.activate(proj_file) do
-                TestEnv.activate() do
-                    _runtests_in_current_env(shouldrun, paths, proj_file, nworkers, nworker_threads, worker_init_expr, testitem_timeout, retries, verbose_results, debug, report, logs)
+            orig_proj = Base.active_project()
```
nickrobinson251, line 217: can you add a comment linking to the issue that this is working around?

nickrobinson251 (approved):
Thanks for looking into this!!

Please open an issue about it in case we can one day figure out why Pkg/TestEnv isn't already doing this? (i.e. i expect `TestEnv.activate` to "see" that if there are no test-only deps, then we should need to re-trigger precompilation)

Also, i'm presuming it's basically impossible to add a test that this avoids the repeated-precompile issue(?), so it'd at least be helpful to have steps for how to reproduce this (i guess in that same issue), in case someone reports seeing this again in the future

src/ReTestItems.jl
```diff
@@ -214,15 +214,27 @@ function _runtests(shouldrun, paths, nworkers::Int, nworker_threads::Int, worker
...
+            try
+                if haskey(TEST_ENVS, proj_file)
+                    testenv = TEST_ENVS[proj_file]
+                    Base.set_active_project(testenv)
```
nickrobinson251, line 221: what happens if the the testenv file has been removed?

nickrobinson251: should we have a `isfile(testenv)` check to be safe?

quinnj: Yeah, good call.

src/ReTestItems.jl
```diff
@@ -214,10 +224,19 @@ function _runtests(shouldrun, paths, nworkers::Int, nworker_threads::Int, worker
...
         else
             @debugv 1 "Activating test environment for `$proj_file`"
-            return Pkg.activate(proj_file) do
-                TestEnv.activate() do
-                    _runtests_in_current_env(shouldrun, paths, proj_file, nworkers, nworker_threads, worker_init_expr, testitem_timeout, retries, verbose_results, debug, report, logs)
+            orig_proj = Base.active_project()
+            try
+                if haskey(TEST_ENVS, proj_file) && isfile(TEST_ENVS[proj_file])
+                    testenv = TEST_ENVS[proj_file]
+                    Base.set_active_project(testenv)
```
NHDaly, line 231: Should we also be checking that the orig_proj hasn't been changed at all? I think we might want to keep a timestamp in the cache as well, and check that the project is still the same as it was before, otherwise this might fail because we need to `instantiate` the test-only project env again.

quinnj: yeah, probably a good idea.

Issue #59, opened 2023-05-22
Have `@testitem` run the test when copy-pasted into REPL


nickrobinson251:

NHDaly:
Fixed in #61, right? :) Thanks!

PR #61, opened 2023-05-22
Run testitem if not in `runtests` call

nickrobinson251:
- #59 

Before (on `main`):
```julia
julia> @testitem "min and max" begin
           @test min(1, 2) == 1
           @test max(1, 2) == 2
       end
TestItem(Base.RefValue{Int64}(0), "min and max", Symbol[], true, Symbol[], 0, "REPL[3]", 1, ".", quote
    #= REPL[3]:2 =#
    #= REPL[3]:2 =# @test min(1, 2) == 1
    #= REPL[3]:3 =#
    #= REPL[3]:3 =# @test max(1, 2) == 2
end, TestSetup[], Base.RefValue{Int64}(0), Test.DefaultTestSet[], Base.RefValue{Int64}(0), ReTestItems.PerfStats[], ReTestItems.ScheduledForEvaluation(false))

julia>
```

After (this PR):
```julia
julia> @testitem "min and max" begin
           @test min(1, 2) == 1
           @test max(1, 2) == 2
       end
18:33:22 | START test item "min and max" at REPL[3]:1
Test Summary: | Pass  Total  Time
min and max   |    2      2  0.0s
18:33:22 | DONE  test item "min and max" <0.1 secs (<0.1% compile), 1.21 K allocs (73.158 KB)

julia> 
```

### DIFF
--- a/src/macros.jl
+++ b/src/macros.jl
@@ -255,8 +255,9 @@ macro testitem(nm, exs...)
         error("expected `@testitem` to have a body")
     end
     q = QuoteNode(exs[end])
+    ti = gensym(:ti)
     esc(quote
-        let ti = $TestItem(
+        let $ti = $TestItem(
             Ref(0), $nm, $tags, $default_imports, $setup, $retries,
             $(String(__source__.file)), $(__source__.line),
             $gettls(:__RE_TEST_PROJECT__, "."),
--- a/src/macros.jl
+++ b/src/macros.jl
@@ -128,6 +128,17 @@ struct TestItem
     stats::Vector{PerfStats} # populated when the test item is finished evaluating
     scheduled_for_evaluation::ScheduledForEvaluation # to keep track of whether the test item has been scheduled for evaluation
 end
+function TestItem(id, name, tags, default_imports, setups, retries, file, line, project_root, code)
+    return TestItem(
+        id, name, tags, default_imports, setups, retries, file, line, project_root, code,
+        TestSetup[],
+        Ref{Int}(0),
+        DefaultTestSet[],
+        Ref{Int}(0),
+        PerfStats[],
+        ScheduledForEvaluation(),
+    )
+end
 
 """
     @testitem "name" [tags=[] setup=[] retries=0 default_imports=true] begin
@@ -211,6 +222,7 @@ macro testitem(nm, exs...)
     retries = 0
     tags = Symbol[]
     setup = Any[]
+    _quote = false  # useful for testing `@testitem` itself
     if length(exs) > 1
         kw_seen = Set{Symbol}()
         for ex in exs[1:end-1]
@@ -231,6 +243,9 @@ macro testitem(nm, exs...)
             elseif kw == :retries
                 retries = ex.args[2]
                 @assert retries isa Integer "`default_imports` keyword must be passed an `Integer`"
+            elseif kw == :_quote
+                _quote = ex.args[2]
+                @assert _quote isa Bool "`_quote` keyword must be passed a `Bool`"
             else
                 error("unknown `@testitem` keyword arg `$(ex.args[1])`")
             end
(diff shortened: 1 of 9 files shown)


nickrobinson251:
In our own tests, we sometimes want the old behaviour, so various tests needed to be updated... those changes feel a little messy, so advice on those particularly appreciated

src/macros.jl
```diff
@@ -128,6 +128,17 @@ struct TestItem
...
+function TestItem(id, name, tags, default_imports, setups, retries, file, line, project_root, code)
+    return TestItem(
+        id, name, tags, default_imports, setups, retries, file, line, project_root, code,
+        TestSetup[],
+        Ref{Int}(0),
+        DefaultTestSet[],
+        Ref{Int}(0),
+        PerfStats[],
+        ScheduledForEvaluation(),
+    )
```
nickrobinson251, line 140: since these fields are always set the same, i thought it made sense to have a simpler constructor; it makes manually constructing a `TestItem` in the tests a lot easier

src/macros.jl
```diff
@@ -241,20 +256,20 @@ macro testitem(nm, exs...)
...
+            $(String(__source__.file)), $(__source__.line),
+            $gettls(:__RE_TEST_PROJECT__, "."),
+            $q,
         )
+            if $_quote || $gettls(:__RE_TEST_RUNNING__, false)::$Bool
+                $store_test_item_setup(ti)
+                ti
+            else # We are not in a `runtests` call, so we run the testitem eagerly.
+                $runtestitem(ti)
+                nothing
```
nickrobinson251, line 270: this is the main change

btw, how do people usually assign to variables in a macro if escaping the whole thing? i've used a `let` block here but idk what best practice is... (should we change to just `esc`ing what's necessary?)

test/_test_log_capture.jl
```diff
@@ -5,7 +5,7 @@ const log_display = Symbol(ENV["LOG_DISPLAY"])
 @testset "log capture logs=$(repr(log_display))" begin
     @testset "TestItem" begin
         @testset "log capture for println" begin
-            ti = @testitem "uses println" begin
+            ti = @testitem "uses println" _quote=true begin
```
nickrobinson251, line 8: this `_quote` keyword is how we now avoid running a `@testitem` immediately when outside of `runtests`. i don't think _users_ should ever need this (only developers of the package)

test/macros.jl
```diff
@@ -1,9 +1,23 @@
...
 
+function expand(f)
+    old = get(task_local_storage(), :__RE_TEST_RUNNING__, false)
+    try
+        task_local_storage()[:__RE_TEST_RUNNING__] = true
+        return f()
+    finally
+        task_local_storage()[:__RE_TEST_RUNNING__] = old
+    end
+end
```
nickrobinson251, line 19: we could have used `_quote` for all these too, but i wanted to make sure the `__RE_TEST_RUNNING__` variable was being respected, so chose this approach here

maybe this should just be its own test and we should use `_quote` everywhere else?

test/macros.jl
```diff
@@ -106,6 +134,10 @@ end
     )
 end
 
+@testset "testitem macro runs immediately outside `runtests`" begin
+    @test nothing == @testitem "run" begin; @test true; end
+end
```
nickrobinson251, line 139: wasn't sure how best to test the new behaviour... suggestions for improvements welcome!

src/macros.jl
```diff
@@ -211,6 +222,7 @@ macro testitem(nm, exs...)
     retries = 0
     tags = Symbol[]
     setup = Any[]
+    _quote = false
```
NHDaly, line 225: Maybe put the explanation comment here in addition to / instead of in the definition?

src/macros.jl
NHDaly: I think the main approach is to manually gensym a variable name outside the `esc`aped expression, and then interpolate that in everywhere.. it's a bit ugly. Like you point out, top-level escape can get ugly fast.

A let-block seems like a prudent approach (and probably a good idea anyway to scope the object nicely).. I guess the main concern would be if inside any user-provided expressions we might actually capture `ti` if the user used a variable of the same name.

Like if they have a global variable called `ti = 5` and they try to use it to name their testitem `@testitem "run-$ti"`, then it would end up using your let block instead! 😮 

So I think probably you _should_ gensym a new name for `ti` and use that, or switch to narrow escapes.

NHDaly: e.g. `ti = gensym("ti")` and then you use `$ti` everywhere in the quote.

src/macros.jl
```diff
@@ -241,20 +256,20 @@ macro testitem(nm, exs...)
...
+            Ref(0), $nm, $tags, $default_imports, $setup, $retries,
+            $(String(__source__.file)), $(__source__.line),
+            $gettls(:__RE_TEST_PROJECT__, "."),
+            $q,
         )
+            if $_quote || $gettls(:__RE_TEST_RUNNING__, false)::$Bool
+                $store_test_item_setup(ti)
+                ti
+            else # We are not in a `runtests` call, so we run the testitem eagerly.
+                $runtestitem(ti)
```
NHDaly, line 269: Will this already default to the right eager logs mode and all the other normal REPL-experience settings?

test/macros.jl
NHDaly: I guess you could use Mocking to ensure that the expected testing functions inside ReTestItems runtestitem were called?

Or you could capture the output and test that it printed a test summary at the end?

NHDaly (commented):
Cool! Thanks for picking this up! :)

src/macros.jl
nickrobinson251: _Great_ question! The logs behaviour is correct but the `verbose_results` one isn't! will fix and add tests

src/macros.jl
```diff
@@ -211,6 +223,7 @@ macro testitem(nm, exs...)
     retries = 0
     tags = Symbol[]
     setup = Any[]
+    _run = true  # useful for testing `@testitem` itself
```
quinnj, line 226: This is a great way to manage this; good idea!

quinnj (approved):

PR #62, opened 2023-05-22
feat: add junit_record! for two test suites

mcmcgrath13:
For `RAITest` it makes sense to collapse multiple `JUnitTestSuite` into one. This PR adds a method for recording one test suite to another.


nickrobinson251 (approved):
thanks!

PR #65, opened 2023-05-23
Store task-local variable when running a testitem

nickrobinson251:
this allows you to identify if you're already in a testitem. this is a bit of a hack to support a use-case where we want a macro `@test_foo` to create a `@testset` if we are already in a `@testitem`, and otherwise create a `@testitem`.

### DIFF
--- a/src/ReTestItems.jl
+++ b/src/ReTestItems.jl
@@ -682,6 +682,10 @@ function runtestitem(ti::TestItem, ctx::TestContext; verbose_results::Bool=false
         end
     end
     Test.push_testset(ts)
+    # This allows us to identify if the code is running inside a `@testitem`, which is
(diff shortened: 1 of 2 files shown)


quinnj (approved):

src/ReTestItems.jl
```diff
@@ -682,6 +682,8 @@ function runtestitem(ti::TestItem, ctx::TestContext; verbose_results::Bool=false
         end
     end
     Test.push_testset(ts)
+    prev = get(task_local_storage(), :__TESTITEM_ACTIVE__, false)
+    task_local_storage()[:__TESTITEM_ACTIVE__] = true
```
NHDaly, line 686: Can you add a comment maybe explaining a tiny bit about _why_ we might want this hack?

NHDaly (approved):
LGTM thanks nick.

PR #67, opened 2023-05-24
Store task-local variable when running a testsetup

nickrobinson251:


quinnj:
Hmmmm, maybe I'm misunderstanding what this was for the other day; wouldn't we want to inject this task-local var _into_ the `@testitem` module that we eval? Why would we want this defined on the _coordinator_ vs. _inside_ the `@testitem` module?

PR #73, opened 2023-06-02
Support `_source` for manually setting location

nickrobinson251:
Similar in spirit to https://github.com/JuliaTesting/ReTestItems.jl/pull/65: Useful for if you've a macro the expands to a `@testitem` and you want to pass in where call-site location, so that you get accurate _call site_ file/line info (rather than the macro definition site)
```
START test item "TI2" at test/callsite.jl:42
```

as a silly example:
```julia
macro foo_test(name)
    _source = QuoteNode(__source__)
    quote
        @testitem $name _source=$_source begin
            @test true
        end
    end
end
```

### DIFF
--- a/test/runtests.jl
+++ b/test/runtests.jl
@@ -1,5 +1,15 @@
 using ReTestItems, Test, Pkg
 
+# Used by test/macros.jl, but macros must be defined at global scope (outside `@testset).
+macro foo_test(name)
+    _source = QuoteNode(__source__)
+    quote
+        @testitem $name _source=$_source _run=false begin
+            @test true
+        end
+    end
+end
+
 @testset "ReTestItems" verbose=true begin
     # track all workers every created
     ALL_WORKERS = []
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "ReTestItems"
 uuid = "817f1d60-ba6b-4fd5-9520-3cf149f6a823"
-version = "1.5.1"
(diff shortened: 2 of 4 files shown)


quinnj (approved):
LGTM, though tests are failing

nickrobinson251:
> tests are failing

oops -- macro definition in the tests wasn't at top level, just added an `@eval` to it

test/runtests.jl
```diff
@@ -1,5 +1,15 @@
...
 
+# Used by test/macros.jl, but macros must be defined at global scope (outside `@testset).
+macro foo_test(name)
+    _source = QuoteNode(__source__)
+    quote
+        @testitem $name _source=$_source _run=false begin
+            @test true
+        end
+    end
+end
```
NHDaly, line 11: For next time: The alternative is that you can put the macro uses inside an `@eval` - that way they won't be macro-lowered until after the macro def has been evaluated.

```julia
julia> @testset "Foo" begin
           @eval macro m() :(2+2) end
           @eval @test @m() == 4
       end
```
but how you have it is nice too

nickrobinson251: that's what i thought, but when i tried that it didn't work
https://github.com/JuliaTesting/ReTestItems.jl/pull/73/commits/71493be6834134db659d75544e924d322a658d67

I guess because i need to `@eval` the `@test` too? why is that? preumably something about not picking up the newly defined name? world-age related? i don't quite understand

PR #75, opened 2023-06-06
Fix weird syntax in tests

c42f:
Found as part of testing for JuliaLang/julia#46372 - JuliaSyntax.jl rejects this weird syntax.  The reference parser allows it, but that kind of seems unintentional: Having a semicolon within `[]` would normally imply `vcat`, but having a semicolon followed by a `for` to make it into a comprehension?  That's just weird and maybe unintended ? :-)


c42f:
Congratulations btw, this is the only project in the general registry which discovered this weird syntax :sweat_smile:

nickrobinson251 (approved):

PR #80, opened 2023-06-13
Move GC run outside of `runtestitem`

### LINKED ISSUE #79: Don't trigger full GC within the time allowed for a test item
Here in `runtestitem` we log the test as `DONE` but then trigger GC _before returning the result_ meaning the worker can timeout and be killed after a test has passed and be marked as failed because GC took too long

https://github.com/JuliaTesting/ReTestItems.jl/blob/c346b2558042e5db966a20d3652baec1f34886b9/src/ReTestItems.jl#L821-L825

nickrobinson251:
- So that GC isn't included in the time allowed for a testitem to run (when a timeout is set). (fix #79)
- Also optimise `convert_results_to_be_transferrable` to further reduce time between test done and it being returned to the coordinator process

~Is `remote_eval`'ing GC in between `remote_eval(w, :(runtestitem(...)))` sufficient for it to guarantee GC will _run_ in between the `runtestitem`, or do we need the coordinator to be `wait`ing on it before the next `runtestitem` to be sure?~ We have to `wait` it seems 😞

### DIFF
--- a/src/ReTestItems.jl
+++ b/src/ReTestItems.jl
@@ -306,7 +306,10 @@ function _runtests_in_current_env(
                         end
                         break
                     end
-                    run_full_gc()
+                    # It takes 2 GCs to do a full mark+sweep
+                    # (the first one is a partial mark, full sweep, the next one is a full mark).
+                    GC.gc(true)
+                    GC.gc(false)
                 end
             end
         elseif !isempty(testitems.testitems)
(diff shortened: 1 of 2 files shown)


nickrobinson251:
hmmm, changing to `remote_fetch` does not seem to have gone well 🤔

Drvi:
Is `run_full_gc` defined on the workers?

nickrobinson251:
ooo good thought

src/ReTestItems.jl
```diff
@@ -405,15 +409,18 @@ function start_and_manage_worker(
             try
                 # if we get a WorkerTerminatedException or TimeoutException
                 # then wait will throw here and we fall through to the outer try-catch
-                @debugv 2 "Waiting on test item result"
+                @debugv 2 "Waiting on result for test item $(repr(testitem.name))"
                 testitem_result = @lock cond wait(cond)
-                @debugv 2 "Recieved test item result"
+                @debugv 2 "Recieved result for test item $(repr(testitem.name))"
```
Drvi, line 414:
```suggestion
                @debugv 2 "Received result for test item $(repr(testitem.name))"
```

nickrobinson251:
Verified in https://github.com/RelationalAI/raicode/pull/14162 that this works just the same as before (same memory levels, no noticeable slow down)

It doesn't fix https://relationalai.atlassian.net/jira/software/c/projects/RAI/issues/RAI-13569
But it still seems a good idea, so will merge

PR #81, opened 2023-06-13
Spawn GC so `runtestitem` can return sooner

### LINKED ISSUE #79: Don't trigger full GC within the time allowed for a test item

nickrobinson251:
- Alternative to #80 
- fix #79

### DIFF
--- a/src/ReTestItems.jl
+++ b/src/ReTestItems.jl
@@ -818,11 +818,16 @@ function runtestitem(
     @debugv 1 "Test item $(repr(name)) done$(_on_worker())."
     push!(ti.testsets, ts)
     push!(ti.stats, stats)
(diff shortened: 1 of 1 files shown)


nickrobinson251:
closing in favour of https://github.com/JuliaTesting/ReTestItems.jl/pull/80 (which seems the safer approach, see https://github.com/JuliaTesting/ReTestItems.jl/issues/79#issuecomment-1588281437)

PR #83, opened 2023-06-14
Replace Condition with Channel

nickrobinson251:
- fix https://github.com/JuliaTesting/ReTestItems.jl/issues/84

The previous implementation was racy, because the testitem run could complete and `notify` the Condition before we had started to `wait` on the Condition, in which case when we did then `wait` we would be waiting for the _next_ notify which would be from the `Timer`


quinnj (approved):
Beauty

PR #85, opened 2023-06-15
Use a Channel to safely populate setups Dict

nickrobinson251:
Spawn a task to populate the shared `Dict` of test-setups, rather than writing to a shared `Dict` in parallel which can result in data being lost. 

```julia
julia> D = Dict{Int,Int}()
Dict{Int64, Int64}()

julia> @sync for i in 1:100_000
           Threads.@spawn $D[$i] = $i
       end

julia> length(D)
99377
```

Before, we would 
- Spawn a bunch of tasks to include files in parallel, with all of them sharing the same reference to setups **Dict**:   
  - https://github.com/JuliaTesting/ReTestItems.jl/blob/b1f3a66d50f41ac1ed8f908bb3cc03c39a86e2a1/src/ReTestItems.jl#L598-L608
- During the include, we eval the testsetup macro which calls `store_test_item_setup`
  - https://github.com/JuliaTesting/ReTestItems.jl/blob/f5c6c9133448a2a2f7055f12878825354dbb98c4/src/macros.jl#L56-L57
- In that function, we populate the shared setups **Dict** (without locking)
  - https://github.com/JuliaTesting/ReTestItems.jl/blob/f5c6c9133448a2a2f7055f12878825354dbb98c4/src/macros.jl#L293

Now, we
- Spawn a bunch of tasks to include files in parallel, with all of them sharing the same reference to setups **Channel**
- (as before) During the include, we eval the testsetup macro which calls `store_test_item_setup` 
- In that function, we put the setups on the shared **Channel**
- ALSO Spawn a "setup" task that takes items of the Channel and populates a Dict
- after the `@sync` block with all files-include tasks has completed, we `close` the Channel and `fetch` the **Dict** from the "setup" task

Also splits `store_test_item_setup` into `store_test_item` and `store_test_setup` since really they're different functions.

Thanks to @Drvi for helping figure this one out

### DIFF
--- a/src/ReTestItems.jl
+++ b/src/ReTestItems.jl
@@ -561,7 +561,7 @@ function include_testfiles!(project_name, projectfile, paths, shouldrun, report:
     subproject_root = nothing  # don't recurse into directories with their own Project.toml.
     root_node = DirNode(project_name; report, verbose=true)
     dir_nodes = Dict{String, DirNode}()
-    # setup_channel is populated in store_test_item_setup when we expand a @testsetup
+    # setup_channel is populated in store_test_setup when we expand a @testsetup
     # we set it below in tls as __RE_TEST_SETUPS__ for each included file
     setup_channel = Channel{Pair{Symbol, TestSetup}}(Inf)
     setup_task = @spawn begin
--- a/src/ReTestItems.jl
+++ b/src/ReTestItems.jl
@@ -561,9 +561,19 @@ function include_testfiles!(project_name, projectfile, paths, shouldrun, report:
     subproject_root = nothing  # don't recurse into directories with their own Project.toml.
     root_node = DirNode(project_name; report, verbose=true)
     dir_nodes = Dict{String, DirNode}()
-    # setups is populated in store_test_item_setup when we expand a @testsetup
+    # setup_channel is populated in store_test_item_setup when we expand a @testsetup
     # we set it below in tls as __RE_TEST_SETUPS__ for each included file
-    setups = Dict{Symbol, TestSetup}()
+    setup_channel = Channel{Pair{Symbol, TestSetup}}(Inf)
+    setup_task = @spawn begin
+        setups = Dict{Symbol, TestSetup}()
+        for (name, setup) in setup_channel
+            if haskey(setups, name)
+                @warn "Encountered duplicate @testsetup with name: `$name`. Replacing..."
+            end
+            setups[name] = setup
+        end
+        return setups
+    end
     hidden_re = r"\.\w"
     @sync for (root, d, files) in Base.walkdir(project_root)
         if subproject_root !== nothing && startswith(root, subproject_root)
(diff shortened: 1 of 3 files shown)


src/ReTestItems.jl
```diff
@@ -561,9 +561,19 @@ function include_testfiles!(project_name, projectfile, paths, shouldrun, report:
     subproject_root = nothing  # don't recurse into directories with their own Project.toml.
     root_node = DirNode(project_name; report, verbose=true)
     dir_nodes = Dict{String, DirNode}()
-    # setups is populated in store_test_item_setup when we expand a @testsetup
+    # setup_channel is populated in store_test_setup when we expand a @testsetup
     # we set it below in tls as __RE_TEST_SETUPS__ for each included file
-    setups = Dict{Symbol, TestSetup}()
+    setup_channel = Channel{Pair{Symbol, TestSetup}}(Inf)
+    setup_task = @spawn begin
```
quinnj, line 567: So it seems like we could possibly use the Channel-bound-to-a-task functionality that I've never actually used here? i.e. keep `setups` being created outside, then do a Channel-task to iterate until empty and then we don't need the `fetch` down below?

This might be a very naive comment and even if the channel-task thing can work, I really don't mind keeping things as-is. It's just I've never actually used the Channel-task-bind thing and it feels like this is an opporunity.

quinnj (approved):
Very nice; great catch again @Drvi. I left one comment of something we can maybe try if we want to be fancy, but I also might be way off there.

src/ReTestItems.jl
nickrobinson251: hmm, i'm not sure i quite follow

we could define the setups outside, like
```diff
+    setups = Dict{Symbol, TestSetup}()
     setup_task = @spawn begin
-        setups = Dict{Symbol, TestSetup}()
         for (name, setup) in setup_channel
-            if haskey(setups, name)
+            if haskey($setups, name)
                 @warn "Encountered duplicate @testsetup with name: `$name`. Replacing..."
             end
-            setups[name] = setup
+            $setups[name] = setup
         end
         return setups
     end
```
which would save us the `fetch` later on 

But i'm not sure i see how the `bind` idea relates

Are you thinking we'd `bind(setup_channel, setup_task)`? As i understand it, that would close the channel when the task terminates. But here, because the task is iterating the channel, we need the opposite: close the channel in order for the task to terminate 

Not super familiar with this so i could well be missing  something!

p.s. I _slightly_ preferred creating the Dict in the task and having to `fetch` it, as it seemed more explicit and one less bit of external state being mutated by a task (which the package already has a lot of and i sometimes find hard to track)

NHDaly: I think Jacob specifically means something like:
````julia
julia> put_ch = Channel() do ch
           for v in ch
               println("got: ", v)
           end
       end
Channel{Any}(0) (empty)

julia> put!(put_ch, 10);
got: 10

julia> put!(put_ch, 100);
got: 100

julia> close(put_ch)

```
but i'm not sure yet how this would be used here; i haven't read anything yet

PR #88, opened 2023-06-21
Record errors before logs in XML reports

nickrobinson251:
for anywhere that truncates the output (such as DataDog), we need to make sure the most important info is high up

fix https://relationalai.atlassian.net/jira/software/c/projects/RAI/issues/RAI-13805


Drvi (approved):
LGTM. Can you update the docstring?

PR #91, opened 2023-06-29
Log at error level when worker killed + simplify retry logic

nickrobinson251:
- Log worker killed (due to crash or timeout) as an error-level log message, to make these cases easier to find in the logs, especially as time-outs and potentially even crashes might not have logs containing "error" (nor "fail")
- Make the log messages for retries in the `nworkers=0` and `nworker>0` cases match, simplifying the logic in the process.

### DIFF
--- a/src/ReTestItems.jl
+++ b/src/ReTestItems.jl
@@ -301,7 +301,7 @@ function _runtests_in_current_env(
                     # (the first one is a partial mark, full sweep, the next one is a full mark).
                     GC.gc(true)
                     GC.gc(false)
-                    if ts.anynonpass && run_number != max_runs
+                    if any_non_pass(ts) && run_number != max_runs
                         run_number += 1
                         @info "Retrying $(repr(testitem.name)). Run=$run_number."
                     else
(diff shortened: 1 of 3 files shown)


src/ReTestItems.jl
```diff
@@ -299,7 +299,7 @@ function _runtests_in_current_env(
                     report_empty_testsets(testitem, ts)
                     if ts.anynonpass && nretries < retry_limit
                         nretries += 1
-                        @warn "Test item $(repr(testitem.name)) failed. Retrying. Retry=$nretries."
+                        @info "Retrying $(repr(testitem.name)). Run=$run_number."
```
nickrobinson251, line 302:
```suggestion
                        @info "Retrying $(repr(testitem.name)). Retry=$nretries."
```

src/ReTestItems.jl
```diff
@@ -290,26 +290,23 @@ function _runtests_in_current_env(
...
-                while nretries ≤ retry_limit
+                run_number = 1
+                max_runs = 1 + max(retries, testitem.retries)
+                while run_number ≤ max_runs
                     res = runtestitem(testitem, ctx; verbose_results, logs)
                     ts = res.testset
-                    print_errors_and_captured_logs(testitem, nretries + 1; logs)
+                    print_errors_and_captured_logs(testitem, run_number; logs)
                     report_empty_testsets(testitem, ts)
-                    if ts.anynonpass && nretries < retry_limit
```
quinnj, line 300: unrelated, but `ts.anynonpass` is awful. every time I see it it makes me think someone is mispelling `anonymous`

PR #94, opened 2023-07-05
Store timeout limit as elapsed time when recording error

nickrobinson251:


src/ReTestItems.jl
```diff
@@ -386,6 +386,7 @@ function record_test_error!(testitem, msg)
     catch e2
         e2 isa TestSetException || rethrow()
     end
+    ts.time_end = ts.time_start + elapsed_seconds
```
mcmcgrath13, line 389: add a comment on why this is needed?

mcmcgrath13 (approved):
thank you!!!

should we be taking advantage of this new feature for any of the other test errors?

nickrobinson251:
we only hit this path for the cases where the worker running the test is killed
which is 2 cases: 
1. we killed the worker because we hit the time out (covered in this PR)
2. the worker crashed as part of running the test

For 2, there's no easy way to know how long the test was running before it crashed (AFAIK), so in this PR is will now set the elapsed time as `0` whereas before this PR it was set to however long it took for this code to create with error (which is roughly a few hundred ms) but that's not a useful number. Is knowing how long the test was running before the crash useful? If so, we could maybe figure out a way to do that... so if we think it'd be useful i will open an issue for that as a follow-up

nickrobinson251:
Opened https://github.com/JuliaTesting/ReTestItems.jl/issues/95 :)

Thanks for suggesting it!

PR #99, opened 2023-07-24
If a worker doesn't start successfully then retry

nickrobinson251:
...but wait for the requested number of workers to start successfully before running any tests (i.e. throw before running tests if we failed to start the requested number of workers).

Similarly, allow retry when starting _replacement_ workers (for workers that died whilst running a test)

In other words, before we were assuming workers could only die whilst a test was running whereas now we handle the possibly of workers dying before they start running a test.

This potentially adds a little latency (increasing time-to-first-test), even if all workers start successfully first try, since we know wait on all workers before starting any tests, whereas previously a worker would start running tests as soon as it was ready (not waiting for all workers to be ready). Workers are still started in parallel, so the latency should be minimal. And so long as starting worker processes is very fast (relative to running tests), i.e. so long as `worker_init_expr` isn't something very time consuming, then this latency from synchronizing should hopefully be acceptable to users even if a retry is needed.

### Examples

- Logs if N workers fail to start even on retry:

```julia
julia> runtests(file; nworkers=4, worker_init_expr)
[ Info: Scanning for test items in project `ReTestItems` at paths: /Users/nickr/repos/ReTestItems.jl/test/testfiles/_happy_tests.jl
[ Info: Finished scanning for test items in 0.06 seconds. Scheduling 3 tests on pid 37946 with 4 worker processes and 2 threads per worker.
[ Info: Starting test workers
  Worker 75750:  [ Info: Starting test worker 3 on pid = 75750, with 2 threads
  Worker 75750:  signal (6): Abort trap: 6
  Worker 75750:  in expression starting at none:1
  Worker 75750:  __pthread_kill at /usr/lib/system/libsystem_kernel.dylib (unknown line)
┌ Error: Worker(pid=75750, terminated=true, termsignal=6) terminated unexpectedly. Starting new worker (retry 1/2).
└ @ ReTestItems ~/repos/ReTestItems.jl/src/ReTestItems.jl:387
  Worker 75751:  [ Info: Starting test worker 4 on pid = 75751, with 2 threads
  Worker 75749:  [ Info: Starting test worker 2 on pid = 75749, with 2 threads
  Worker 75748:  [ Info: Starting test worker 1 on pid = 75748, with 2 threads
  Worker 75748:  signal (6): Abort trap: 6
  Worker 75748:  in expression starting at none:1
  Worker 75748:  __pthread_kill at /usr/lib/system/libsystem_kernel.dylib (unknown line)
┌ Error: Worker(pid=75748, terminated=true, termsignal=6) terminated unexpectedly. Starting new worker (retry 1/2).
└ @ ReTestItems ~/repos/ReTestItems.jl/src/ReTestItems.jl:387
  Worker 75865:  [ Info: Starting test worker 3 on pid = 75865, with 2 threads
  Worker 75866:  [ Info: Starting test worker 1 on pid = 75866, with 2 threads
  Worker 75866:  signal (6): Abort trap: 6
  Worker 75866:  in expression starting at none:1
┌ Error: Worker(pid=75866, terminated=true, termsignal=6) terminated unexpectedly. Starting new worker (retry 2/2).
└ @ ReTestItems ~/repos/ReTestItems.jl/src/ReTestItems.jl:387
  Worker 75970:  [ Info: Starting test worker 1 on pid = 75970, with 2 threads
  Worker 75970:  signal (6): Abort trap: 6
  Worker 75970:  in expression starting at none:1
ERROR: TaskFailedException

    nested task error: ReTestItems.Workers.WorkerTerminatedException(Worker(pid=75970, terminated=true, termsignal=6))
```

- Logs if N workers successfully start given retries:

```julia
julia> runtests(file; nworkers=4, worker_init_expr)
[ Info: Scanning for test items in project `ReTestItems` at paths: /Users/nickr/repos/ReTestItems.jl/test/testfiles/_happy_tests.jl
[ Info: Finished scanning for test items in 0.05 seconds. Scheduling 3 tests on pid 37946 with 4 worker processes and 2 threads per worker.
[ Info: Starting test workers
  Worker 88298:  [ Info: Starting test worker 1 on pid = 88298, with 2 threads
  Worker 88299:  [ Info: Starting test worker 2 on pid = 88299, with 2 threads
  Worker 88298:  signal (6): Abort trap: 6
  Worker 88298:  in expression starting at none:1
  Worker 88298:  __pthread_kill at /usr/lib/system/libsystem_kernel.dylib (unknown line)
┌ Error: Worker(pid=88298, terminated=true, termsignal=6) terminated unexpectedly. Starting new worker (retry 1/2).
└ @ ReTestItems ~/repos/ReTestItems.jl/src/ReTestItems.jl:387
  Worker 88301:  [ Info: Starting test worker 4 on pid = 88301, with 2 threads
  Worker 88300:  [ Info: Starting test worker 3 on pid = 88300, with 2 threads
  Worker 88404:  [ Info: Starting test worker 1 on pid = 88404, with 2 threads
[ Info: Starting evaluating test items
  Worker 88300:  16:59:11 | START (3/3) test item "happy 3" at test/testfiles/_happy_tests.jl:9
  Worker 88299:  16:59:11 | START (2/3) test item "happy 2" at test/testfiles/_happy_tests.jl:5
  Worker 88404:  16:59:11 | START (1/3) test item "happy 1" at test/testfiles/_happy_tests.jl:1
  Worker 88300:  16:59:11 | DONE  (3/3) test item "happy 3" <0.1 secs (93.4% compile), 69.42 K allocs (4.200 MB)
  Worker 88404:  16:59:11 | DONE  (1/3) test item "happy 1" <0.1 secs (94.1% compile), 69.42 K allocs (4.200 MB)
  Worker 88299:  16:59:11 | DONE  (2/3) test item "happy 2" <0.1 secs (93.9% compile), 69.42 K allocs (4.200 MB)
Test Summary:                        | Pass  Total  Time
ReTestItems                          |    3      3  3.2s
  test                               |    3      3
    test/testfiles                   |    3      3
      test/testfiles/_happy_tests.jl |    3      3
        happy 1                      |    1      1  0.1s
        happy 2                      |    1      1  0.1s
        happy 3                      |    1      1  0.1s
```

### DIFF
--- a/src/ReTestItems.jl
+++ b/src/ReTestItems.jl
@@ -371,7 +371,7 @@ end
 # Want to be somewhat robust to workers possibly terminating during start up (e.g. due to
 # the `worker_init_expr`).
 # The number of retries and delay between retries is currently arbitrary...
-# we want to retry at least once, and we give a slight delay in case their are resources
+# we want to retry at least once, and we give a slight delay in case there are resources
 # that need to be cleaned up before a new worker would be able to start successfully.
 const _NRETRIES = 2
 const _RETRY_DELAY_SECONDS = 1
--- a/src/ReTestItems.jl
+++ b/src/ReTestItems.jl
@@ -311,16 +311,30 @@ function _runtests_in_current_env(
                 end
             end
         elseif !isempty(testitems.testitems)
-            # spawn a task per worker to start and manage the lifetime of the worker
-            # get starting test items for each worker
+            # Use the logger that was set before we eval'd any user code to avoid world age
+            # issues when logging https://github.com/JuliaLang/julia/issues/33865
+            original_logger = current_logger()
+            # Wait for all workers to be started so we can throw as soon as possible if
+            # there was an error starting a worker.
+            @info "Starting test workers"
+            workers = Vector{Worker}(undef, nworkers)
+            ntestitems = length(testitems.testitems)
+            @sync for i in 1:nworkers
+                @spawn begin
+                    with_logger(original_logger) do
+                        $workers[$i] = start_worker($proj_name, $nworker_threads, $worker_init_expr, $ntestitems)
+                    end
+                end
+            end
+            # Now all workers are started, we can begin processing test items.
+            @info "Starting evaluating test items"
             starting = get_starting_testitems(testitems, nworkers)
-            @sync for i = 1:nworkers
+            @sync for i in 1:nworkers
+                w = workers[i]
                 ti = starting[i]
                 @spawn begin
-                    # Wrapping with the logger that was set before we eval'd any user code to
-                    # avoid world age issues when logging https://github.com/JuliaLang/julia/issues/33865
-                    with_logger(current_logger()) do
-                        start_and_manage_worker($proj_name, $testitems, $ti, $nworker_threads, $worker_init_expr, $testitem_timeout, $retries, $verbose_results, $debug, $report, $logs)
+                    with_logger(original_logger) do
+                        manage_worker($w, $proj_name, $testitems, $ti, $nworker_threads, $worker_init_expr, $testitem_timeout, $retries, $verbose_results, $debug, $report, $logs)
                     end
                 end
             end
@@ -345,7 +359,7 @@ function start_worker(proj_name, nworker_threads, worker_init_expr, ntestitems)
         Test.TESTSET_PRINT_ENABLE[] = false
         const GLOBAL_TEST_CONTEXT = ReTestItems.TestContext($proj_name, $ntestitems)
         GLOBAL_TEST_CONTEXT.setups_evaled = ReTestItems.TestSetupModules()
-        @info "Starting test item evaluations on pid = $(Libc.getpid()), with $(Threads.nthreads()) threads"
+        @info "Starting test worker on pid = $(Libc.getpid()), with $(Threads.nthreads()) threads"
         $(worker_init_expr.args...)
         nothing
     end)
@@ -396,13 +410,12 @@ function record_test_error!(testitem, msg, elapsed_seconds::Real=0.0)
     return testitem
 end
 
-function start_and_manage_worker(
-    proj_name, testitems, testitem, nworker_threads, worker_init_expr,
+function manage_worker(
+    worker::Worker, proj_name, testitems, testitem, nworker_threads, worker_init_expr,
     timeout::Real, retries::Int, verbose_results::Bool, debug::Int, report::Bool, logs::Symbol
 )
     ntestitems = length(testitems.testitems)
-    worker = start_worker(proj_name, nworker_threads, worker_init_expr, ntestitems)
-    run_number = 1
+    @show run_number = 1
     while testitem !== nothing
         ch = Channel{TestItemResult}(1)
         testitem.workerid[] = worker.pid
--- a/src/ReTestItems.jl
+++ b/src/ReTestItems.jl
@@ -315,22 +315,21 @@ function _runtests_in_current_env(
             # issues when logging https://github.com/JuliaLang/julia/issues/33865
             original_logger = current_logger()
             # Wait for all workers to be started so we can throw as soon as possible if
-            # there was an error starting a worker.
+            # we were unable to start the requested number of workers
             @info "Starting test workers"
             workers = Vector{Worker}(undef, nworkers)
             ntestitems = length(testitems.testitems)
             @sync for i in 1:nworkers
                 @spawn begin
                     with_logger(original_logger) do
-                        $workers[$i] = start_worker($proj_name, $nworker_threads, $worker_init_expr, $ntestitems)
+                        $workers[$i] = robust_start_worker($proj_name, $nworker_threads, $worker_init_expr, $ntestitems; worker_num=$i)
                     end
                 end
             end
             # Now all workers are started, we can begin processing test items.
             @info "Starting evaluating test items"
             starting = get_starting_testitems(testitems, nworkers)
-            @sync for i in 1:nworkers
-                w = workers[i]
+            @sync for (i, w) in enumerate(worker)
                 ti = starting[i]
                 @spawn begin
                     with_logger(original_logger) do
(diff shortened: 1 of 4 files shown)


src/ReTestItems.jl
```diff
@@ -395,13 +438,11 @@ function record_test_error!(testitem, msg, elapsed_seconds::Real=0.0)
...
 
-function start_and_manage_worker(
-    proj_name, testitems, testitem, nworker_threads, worker_init_expr,
+function manage_worker(
+    worker::Worker, proj_name, testitems, testitem, nworker_threads, worker_init_expr,
     timeout::Real, retries::Int, verbose_results::Bool, debug::Int, report::Bool, logs::Symbol
 )
     ntestitems = length(testitems.testitems)
-    worker = start_worker(proj_name, nworker_threads, worker_init_expr, ntestitems)
-    run_number = 1
```
Drvi, line 404: I think the rest of this function relies on `run_number` being defined, e.g. the `print_errors_and_captured_logs` call and below

src/ReTestItems.jl
```diff
@@ -337,21 +350,51 @@ function _runtests_in_current_env(
...
         using ReTestItems, Test
         Test.TESTSET_PRINT_ENABLE[] = false
         const GLOBAL_TEST_CONTEXT = ReTestItems.TestContext($proj_name, $ntestitems)
         GLOBAL_TEST_CONTEXT.setups_evaled = ReTestItems.TestSetupModules()
-        @info "Starting test item evaluations on pid = $(Libc.getpid()), with $(Threads.nthreads()) threads"
+        @info "Starting test worker$($i) on pid = $(Libc.getpid()), with $(Threads.nthreads()) threads"
         $(worker_init_expr.args...)
         nothing
     end)
     return w
```
Drvi, line 368: My theory is that there is a race between the internal tasks we spawn in `Worker` and the lethal `worker_init_expr` here. So I tried to running the expression in a separate `remote_fetch`, after the we know everything has been setup for the previous `remote_fetch`, and now I'm running `"worker crashes immediately"` test in a while loop for a couple of minutes without a hang.
```suggestion
        nothing
    end)
    remote_fetch(w, worker_init_expr)
    return w
```

Drvi: Nope, I was not interpolating the expression correctly... sorry for the noise

Drvi:
It seems like the cause of the hangs is `IOCapture`, somehow we never trigger whatever makes the [async task](https://github.com/JuliaDocs/IOCapture.jl/blob/a23f2a4e0b48a7170c040a5f2a82abc5ba0757bb/src/IOCapture.jl#L108) finish when we kill the workers abruptly, I don't understand why, but we are not the only ones getting stuck at that point: https://github.com/JuliaDocs/IOCapture.jl/issues/17

When I rewrite the testcase like this:
```julia
@testset "worker crashes immediately" begin
    file = joinpath(TEST_FILES_DIR, "_happy_tests.jl")
    worker_init_expr = quote
        @eval ccall(:abort, Cvoid, ())
    end
    mktemp() do io, path
        redirect_stdio(stdout=io, stderr=io, stdin=devnull) do
            encased_testset() do
                runtests(file; nworkers=4, worker_init_expr)
            end
        end
        @show read(path, String)
    end
end
```
the hangs don't happen anymore.

I did notice one issue though: the `redirect_worker_output` task can end without reading all available bytes from the worker process, I suggest adding
```julia
    finally
        sleep(0.1)
        bytesavailable(proc) > 0 && (fn(io, w.pid, String(Base.readavailable(proc))); flush(io))
```
to the try catch block (I've seen that `bytesavailable(proc)` can be > 0 in the finally branch)

nickrobinson251:
> When I rewrite the testcase like this ... the hangs don't happen anymore.

Thanks!

nickrobinson251:
> I did notice one issue though

Is that directly connected to this PR? Could you make a PR with that fix, ideally with a test if you know how to reproduce?

Drvi:
I can make a separate PR, but I don't know how to test it. The issue seems to only happen when running under IOCapture :/

src/ReTestItems.jl
```diff
@@ -396,12 +439,11 @@ function record_test_error!(testitem, msg, elapsed_seconds::Real=0.0)
...
 end
 
-function start_and_manage_worker(
-    proj_name, testitems, testitem, nworker_threads, worker_init_expr,
+function manage_worker(
+    worker::Worker, proj_name, testitems, testitem, nworker_threads, worker_init_expr,
     timeout::Real, retries::Int, verbose_results::Bool, debug::Int, report::Bool, logs::Symbol
 )
     ntestitems = length(testitems.testitems)
-    worker = start_worker(proj_name, nworker_threads, worker_init_expr, ntestitems)
```
Drvi, line 404: Any thoughts about the alternative design where we keep spawning the workers eagerly here, but with `robust_start_worker` so they can start working asap? If any of the workers failed to start at this point, we would throw an exception which would be caught in the caller, `terminate!`ing all the other workers and failing the run (so we'd still require all workers to start successfully for `runtests` to succeed).

test/integrationtests.jl
```diff
@@ -769,4 +769,71 @@ end
...
+        # Test that `runtests` errored overall, before any test items were run.
+        @test n_tests(results) == 1
+        @test length(errors(results)) == 1
+    end
+end
+
+@testset "worker crashes immediately but succeeds on retry" begin
+    file = joinpath(TEST_FILES_DIR, "_happy_tests.jl")
+    nworkers = 4
+    tmpdir = joinpath("/tmp", "JL_RETESTITEMS_TEST_TMPDIR")
```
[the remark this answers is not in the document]
nickrobinson251: there was a reason we couldn't do that ...although i can't remember what it was

src/ReTestItems.jl
nickrobinson251: perhaps i'm not understanding, but i think in that case the runtests would fail _after_ all the successful workers had completed running there tests (i.e. potentially after hours of unnecessary processing)

test/integrationtests.jl
```diff
@@ -769,4 +769,71 @@ end
...
+    tmpfile = joinpath(tmpdir, "worker_crash")
+    # First worker start should crash, then all subsequent worker starts should succeed.
+    worker_init_expr = quote
+        if !isfile($tmpfile)
+            mkpath($tmpdir)
+            write($tmpfile, "1")
+            @eval ccall(:abort, Cvoid, ())
+        end
```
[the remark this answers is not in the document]
nickrobinson251: ah yeah, it's not necessary for the test, but i've improved the comment

test/integrationtests.jl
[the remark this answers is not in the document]
nickrobinson251: added a comment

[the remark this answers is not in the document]
Drvi: Hmm... The `cleanup` functionality should be tied to the calling process (the one that calls `runtests`), not the spawned worker processes, so that should be ok, no? We can also do manual cleanup with `mktempdir(...;cleanup=false)`. Also, `/tmp` would fail on windows, `tmpdir()`  should handle this across platforms.

src/ReTestItems.jl
Drvi: Some workers might be in the middle of running tests, yes, but if we detect that any worker failed to start with retries, we'd call `terminate!` on _all_ workers, so they should be interrupted/killed quickly (like we do when they time-out).

test/integrationtests.jl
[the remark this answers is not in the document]
nickrobinson251: i remember we previously had issues with `tmpdir` so i'm reluctant to go back

i don't think we need be concerned about the tests for ReTestItems running on windows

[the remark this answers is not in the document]
nickrobinson251: ah, i see how the other case of `/tmp` is different -- in this PR i think you're right and `mktemp` is fine

src/ReTestItems.jl
Drvi: > how would we do that?
I guess with a flag that says whether to retry workers or not, but I didn't mean to delve into implementation detail, I wanted to discuss the overall idea. I think in a lot of cases the latency diff won't be noticeable, it might be depending on what `worker_init_expr` is and how many workers one uses. But if you are set on the current approach I still think it's an improvement, don't get me wrong.

test/integrationtests.jl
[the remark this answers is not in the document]
Drvi: Thanks! Yeah I think if we have some contributors using Windows, this would make it a bit easier for them.

Drvi (approved):
LGTM, should be a strict improvement for most use cases.

src/ReTestItems.jl
nickrobinson251: i like how simple the current approach here is, and don't _expect_ the latency to be noticeable, which is why i'd rather wait and see if we need to do anything more

nickrobinson251: thanks for making me think about it more!

quinnj (commented):
Thanks for tackling this @nickrobinson251; looks good to me

PR #100, opened 2023-07-25
Introduce `RETESTITEMS_TIMEOUT` env variable

Drvi:


src/ReTestItems.jl
```diff
@@ -174,7 +175,7 @@ function runtests(
     nworkers::Int=parse(Int, get(ENV, "RETESTITEMS_NWORKERS", "0")),
     nworker_threads::Union{Int,String}=get(ENV, "RETESTITEMS_NWORKER_THREADS", "2"),
     worker_init_expr::Expr=Expr(:block),
-    testitem_timeout::Real=DEFAULT_TEST_ITEM_TIMEOUT,
+    testitem_timeout::Real=parse(Float64, get(ENV, "RETESTITEMS_TIMEOUT", string(DEFAULT_TEST_ITEM_TIMEOUT))),
```
nickrobinson251, line 178: i've always found `testitem_timeout` a verbose name, but i presumed it was because we wanted to reserve space for a possible "global" `timeout` (i.e. all tests in total must finish in X seconds, not just each individual test item)...  

if that's the case, then we should probably do the same with the ENV var, and have it be named something like `RETESTITEMS_TESTITEM_TIMEOUT`

I also think matching the keyword is good for providing a consistent interface

(and we probably want to edit the internal const to be spelled `DEFAULT_TESTITEM_TIMEOUT` for consistency)

```suggestion
    testitem_timeout::Real=parse(Float64, get(ENV, "RETESTITEMS_TESTITEM_TIMEOUT", string(DEFAULT_TESTITEM_TIMEOUT))),
```

src/ReTestItems.jl
```diff
@@ -111,8 +111,9 @@ searched for `@testitem`s.
 will be run.
 
 ## Configuring `runtests`
-- `testitem_timeout::Int`: The number of seconds to wait until a `@testitem` is marked as failed.
-  Defaults to 30 minutes. Note timeouts are currently only applied when `nworkers > 0`.
+- `testitem_timeout::Real`: The number of seconds to wait until a `@testitem` is marked as failed.
+  Defaults to 30 minutes. Can also be set using the `RETESTITEMS_TIMEOUT` environment variable.
```
nickrobinson251, line 115:
```suggestion
  Defaults to 30 minutes. Can also be set using the `RETESTITEMS_TESTITEM_TIMEOUT` environment variable.
```

nickrobinson251 (approved):
Please update the name, otherwise LGTM - thanks!

test/integrationtests.jl
```diff
@@ -659,6 +659,23 @@ end
     @test err.value == string(ErrorException("Timed out after 4s evaluating test item \"Test item takes 60 seconds\" (run=1)"))
 end
 
+@testset "testitem timeout set via env variable" begin
+    file = joinpath(TEST_FILES_DIR, "_timeout_tests.jl")
+    # NOTE: this test must run with exactly 1 worker, so that we can test that the worker
+    # is replaced after the timeout and subsequent testitems still run.
+    nworkers = 1
+    @assert nworkers == 1
+    results = withenv("RETESTITEMS_TIMEOUT" => "4.0") do
```
Drvi, line 668:
```suggestion
    results = withenv("RETESTITEMS_TESTITEM_TIMEOUT" => "4.0") do
```

PR #104, opened 2023-08-01
Check the worker process for any available bytes

Drvi:
before we stop redirecting its output. When printing the `bytesavailable(proc)` from the added code segment it occasionally returned non-zeros (in my experiments where the workers were instantly aborted in the `init_worker_expr`). I don't know how to test this properly, but there seems to be no harm in adding it.

### DIFF
--- a/src/workers.jl
+++ b/src/workers.jl
@@ -206,6 +206,10 @@ function redirect_worker_output(io::IO, w::Worker, fn, proc)
         # @error "Error redirecting worker output $(w.pid)" exception=(e, catch_backtrace())
         terminate!(w, :redirect_worker_output)
         e isa EOFError || e isa Base.IOError || rethrow()
+    finally
+        # Making sure we read all the output from the worker process in case it was suddenly killed.
+        sleep(0.05)
+        bytesavailable(proc) > 0 && (fn(io, w.pid, String(Base.readavailable(proc))); flush(io))
     end
     true
 end


nickrobinson251:
> in my experiments

Can you share how you were experimenting with this? Even if we don't want to add it as a test, it'd be good to know how to go about triggering the issue

nickrobinson251:
Does ConcurrentUtils need the same fix?

Drvi:
>how you were experimenting with this?

I added `display(string("bytes: $(bytesavailable(proc))"))` between the `sleep` and `bytesavailable` rows from this PR and then I was running [this version](https://github.com/JuliaTesting/ReTestItems.jl/blob/c427b022770a1c68b885f4fb1d2a7f28ceba198f/test/integrationtests.jl#L772-L786) of the `"worker crashes immediately"` test. Then very rarely this would happen:
```julia
julia> @testset "worker crashes immediately" begin
           file = joinpath(TEST_FILES_DIR, "_happy_tests.jl")
           worker_init_expr = quote
               if iseven(Libc.getpid())
                   @eval ccall(:abort, Cvoid, ())
                   # sleep(10)
               end
           end
           captured = IOCapture.capture() do
               encased_testset() do
                   runtests(file; nworkers=4, worker_init_expr)
               end
           end
           @show captured
       end

"bytes: 0"
"bytes: 0"
"bytes: 0"
"bytes: 328"
"""bbybytytetess:es:  6:69 9""6

9"
```

>Does ConcurrentUtils need the same fix?

I think so, yeah

nickrobinson251 (approved):

PR #108, opened 2023-08-24
Add ability to pass an expression to run after every testitem

nickrobinson251:
Support passing to `runtests` an expression to run on the worker process immediately after each `@testitem` has been run.

This is useful when you want to guarantee that all the code being exercised by your tests upholds some invariant. For example, suppose you're testing a system that does some manual memory management, for which is crucial that all resources that get marked in-use subsequently get marker as free. We could add to every single test-item a final line `@test everything_got_freed()`, which is probably a good choice for a small enough test suite, but this has a few downsides especially for larger code-bases: (i) it is easy for this line to be forgotten especially when there are many developers who now all need to remember to add this, (ii) it can add up to a lot of extra code when you've many hundreds or thousands of test-items, (iii) it arguably isn't the right place for this test if we want it to run on test-items that are for much higher-level code (that is built some layers up from the manual memory management in our example), even though it might be interesting to run this test after the high-level code (since that high-level code could have combined some utilities provided a middle-layer in an interesting way that exposes an issue with how the low-level memory management code is being used); by keeping the low-level check in the `test_end_expr` we can run it on every test-item without every developer having to be responsible for it.

FYI @li1

### DIFF
--- a/src/ReTestItems.jl
+++ b/src/ReTestItems.jl
@@ -842,13 +842,16 @@ function runtestitem(
     log_testitem_start(ti, ctx.ntestitems)
     ts = DefaultTestSet(name; verbose=verbose_results)
     stats = PerfStats()
-    # start with empty block expr and build up our @testitem module body
+    # start with empty block expr and build up our `@testitem` and `test_end_expr` module bodies
     body = Expr(:block)
+    test_end_body = Expr(:block)
     if ti.default_imports
         push!(body.args, :(using Test))
+        push!(test_end_body.args, :(using Test))
         if !isempty(ctx.projectname)
             # this obviously assumes we're in an environment where projectname is reachable
             push!(body.args, :(using $(Symbol(ctx.projectname))))
+            push!(test_end_body.args, :(using $(Symbol(ctx.projectname))))
         end
     end
     Test.push_testset(ts)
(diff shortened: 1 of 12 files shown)


src/ReTestItems.jl
```diff
@@ -865,6 +868,13 @@ function runtestitem(
             push!(body.args, :(const $setup = $ts_mod))
         end
         @debugv 1 "Setup for test item $(repr(name)) done$(_on_worker())."
+
+        # add the `test_end_expr` to a module to be run after the test item
+        test_end_body = copy(body)
+        test_end_mod_expr = :(module $(gensym(name * " test_end")) end)
+        append!(test_end_body.args, test_end_expr.args)
+        test_end_mod_expr.args[3] = test_end_body
```
nickrobinson251, line 876: should we also `softscope` the args, like we do for the `@testitem` body (but don't for the `worker_init_expr`)?

test/integrationtests.jl
```diff
@@ -827,4 +827,115 @@ end
...
+            (isnothing(p) || isempty(p.pages)) && return nothing
+            @testset "no pins left at end of test" begin
+                @test count_pins(p) == 0
+            end
+        end
+        results_with_end = with_test_package("TestEndExpr.jl") do
+            runtests(; nworkers=1, worker_init_expr, test_end_expr)
+        end
+        @test !all_passed(results_with_end)
+        @test length(failures(results_with_end)) ≥ 1
```
nickrobinson251, line 937:
```suggestion
        @test n_passed(results_with_end) ≥ 1
        @test length(failures(results_with_end)) ≥ 1
```

src/ReTestItems.jl
NHDaly: That seems reasonable to me, yeah

nickrobinson251: done

src/ReTestItems.jl
```diff
@@ -245,7 +258,7 @@ function _runtests(shouldrun, paths, nworkers::Int, nworker_threads::String, wor
         if is_running_test_runtests_jl(proj_file)
             # Assume this is `Pkg.test`, so test env already active.
             @debugv 2 "Running in current environment `$(Base.active_project())`"
-            return _runtests_in_current_env(shouldrun, paths, proj_file, nworkers, nworker_threads, worker_init_expr, testitem_timeout, retries, verbose_results, debug, report, logs)
+            return _runtests_in_current_env(shouldrun, paths, proj_file, nworkers, nworker_threads, worker_init_expr, test_end_expr, testitem_timeout, retries, verbose_results, debug, report, logs)
         else
```
NHDaly, line 262: 👀 any idea why this line isn't covered?

nickrobinson251: maybe fixed by latest commit -- but i think this is just CodeCov being confusing

this line is exercised whenever a package using ReTestItems.jl has test triggered via `Pkg.test()`, so it is covered by existing tests for our existing `test/packages/`, but in the latest commit i add the new TestEndExpr.jl package as well to make sure that works with `Pkg.test()` too

src/ReTestItems.jl
```diff
@@ -865,17 +878,18 @@ function runtestitem(
             push!(body.args, :(const $setup = $ts_mod))
         end
         @debugv 1 "Setup for test item $(repr(name)) done$(_on_worker())."
+
+        # add the `test_end_expr` to a module to be run after the test item
+        test_end_body = copy(body)
```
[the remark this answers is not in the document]
quinnj: Hmmm....I guess that gets a little tricky, because the `test_end_expr` can't really express that it needs test setups apart from the test items. It doesn't feel like we should auto-import setups into the test_end_expr though; like, the `test_end_expr` should be stand-alone and any imports you need, you need to put them there explicitly.

quinnj (commented):
This looks pretty good to me; left one question on test setups for the test_end_expr. I also couldn't decipher why the macos test had 2 failures?

NHDaly (approved):
LGTM! Pretty clear and straightforward, including the tests. thanks nick! :)

src/ReTestItems.jl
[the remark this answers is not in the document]
nickrobinson251: ah, good catch. i just did `copy` to save duplicating code, but all i wanted was to get `using Test` and `using PackageBeingTested` included -- i can move this to not also import the test-setups

nickrobinson251:
> why the macos test had 2 failures?

I think flaky test -- will open an issue

src/ReTestItems.jl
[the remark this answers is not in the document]
nickrobinson251: changed to not import test-setups

PR #109, opened 2023-08-30
Replace worker process when memory pressure high

nickrobinson251:
Workaround for workers accumulating memory (due to running in modules, which may not get garbage collected https://github.com/JuliaLang/julia/issues/48711). Allows setting a threshold on memory usage, at which point we start proactively replacing worker processes, to (hopefully) avoid test-items failing due to workers being Out Of Memory (OOM) killed

### DIFF
--- a/src/ReTestItems.jl
+++ b/src/ReTestItems.jl
@@ -16,6 +16,7 @@ export TestSetup, TestItem, TestItemResult
 const RETESTITEMS_TEMP_FOLDER = mkpath(joinpath(tempdir(), "ReTestItemsTempLogsDirectory"))
 const DEFAULT_TESTITEM_TIMEOUT = 30*60
 const DEFAULT_RETRIES = 0
+const DEFAULT_MEMORY_THRESHOLD = 0.99
 
 if isdefined(Base, :errormonitor)
     const errmon = Base.errormonitor
@@ -189,6 +190,7 @@ function runtests(
     worker_init_expr::Expr=Expr(:block),
     testitem_timeout::Real=parse(Float64, get(ENV, "RETESTITEMS_TESTITEM_TIMEOUT", string(DEFAULT_TESTITEM_TIMEOUT))),
     retries::Int=parse(Int, get(ENV, "RETESTITEMS_RETRIES", string(DEFAULT_RETRIES))),
+    memory_threshold::Real=parse(Float64, get(ENV, "RETESTITEMS_MEMORY_THRESHOLD", string(DEFAULT_MEMORY_THRESHOLD))),
     debug=0,
     name::Union{Regex,AbstractString,Nothing}=nothing,
     tags::Union{Symbol,AbstractVector{Symbol},Nothing}=nothing,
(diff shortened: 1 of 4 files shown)


src/ReTestItems.jl
```diff
@@ -16,6 +16,7 @@ export TestSetup, TestItem, TestItemResult
 const RETESTITEMS_TEMP_FOLDER = mkpath(joinpath(tempdir(), "ReTestItemsTempLogsDirectory"))
 const DEFAULT_TESTITEM_TIMEOUT = 30*60
 const DEFAULT_RETRIES = 0
+const DEFAULT_MEMORY_THRESHOLD = 0.99
```
quinnj, line 19: Should we perhaps be more conservative here and set this to `0.9` or `0.95`? It seems like we might still run into issues if, for example, memory was 98% and then we kick off another test item to run.

quinnj (approved):
This LGTM! As discussed, it'd be great to include a test somehow.

src/ReTestItems.jl
nickrobinson251: idk... i was wondering if this should even just not be on by default (i.e. set to 1)

in raicode we should be setting it ourselves anyway

quinnj: Yeah, that's fair. I'd be fine "leaving it off" as `1.0`.

test/integrationtests.jl
```diff
@@ -946,4 +946,49 @@ end
     end
 end
 
+@testset "Replace workers when we hit memory threshold" begin
+    using IOCapture
+    file = joinpath(TEST_FILES_DIR, "_happy_tests.jl")
+    try
+        # monkey-patch the internal `memory_percent` function to return a fixed value, so we
+        # can control if we hit the `memory_threshold`.
+        @eval ReTestItems.memory_percent() = 83.1
```
quinnj, line 955: Beautimus

src/ReTestItems.jl
```diff
@@ -454,13 +465,20 @@ function record_test_error!(testitem, msg, elapsed_seconds::Real=0.0)
...
+    worker::Worker, proj_name::AbstractString, testitems::TestItems, testitem::Union{TestItem,Nothing}, nworker_threads, worker_init_expr::Expr, test_end_expr::Expr,
+    timeout::Real, retries::Int, memory_threshold::Real, verbose_results::Bool, debug::Int, report::Bool, logs::Symbol
 )
     ntestitems = length(testitems.testitems)
     run_number = 1
+    memory_threshold_percent = 100*memory_threshold
     while testitem !== nothing
         ch = Channel{TestItemResult}(1)
+        if memory_percent() > memory_threshold_percent
+            @warn "Memory usage ($(Base.Ryu.writefixed(memory_percent(), 1))%) is higher than limit ($(Base.Ryu.writefixed(memory_threshold_percent, 1))%). Restarting worker process to try to free memory."
```
nickrobinson251, line 477:
```suggestion
            @warn "Memory usage ($(Base.Ryu.writefixed(memory_percent(), 1))%) is higher than threshold ($(Base.Ryu.writefixed(memory_threshold_percent, 1))%). Restarting worker process to try to free memory."
```

src/ReTestItems.jl
nickrobinson251: i think this is probably useful enough and 99% conservative enough that i'll leave this as the default for now -- we could always change it later

PR #111, opened 2023-10-03
Allow setting the `logs` argument via the environment

kpamnany:
With `RETESTITEMS_LOGS` set to the appropriate string ("batched", "eager" or "issues").

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "ReTestItems"
 uuid = "817f1d60-ba6b-4fd5-9520-3cf149f6a823"
-version = "1.18.1"
+version = "1.19.0"
 
 [deps]
 Dates = "ade2ca70-3891-5945-98fb-dc099432e06a"
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "ReTestItems"
 uuid = "817f1d60-ba6b-4fd5-9520-3cf149f6a823"
-version = "1.18.0"
+version = "1.18.1"
 
 [deps]
 Dates = "ade2ca70-3891-5945-98fb-dc099432e06a"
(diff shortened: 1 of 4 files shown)


src/log_capture.jl
```diff
@@ -23,12 +23,12 @@ function default_log_display_mode(report::Bool, nworkers::Integer, interactive::
...
         if report || nworkers > 1
-            return :batched
+            return "batched"
         else
-            return :eager
+            return "eager"
         end
     else
-        return :issues
+        return "issues"
```
nickrobinson251, line 31: why did this need to change?

src/ReTestItems.jl
```diff
@@ -203,7 +203,7 @@ function runtests(
     name::Union{Regex,AbstractString,Nothing}=nothing,
     tags::Union{Symbol,AbstractVector{Symbol},Nothing}=nothing,
     report::Bool=parse(Bool, get(ENV, "RETESTITEMS_REPORT", "false")),
-    logs::Symbol=default_log_display_mode(report, nworkers),
+    logs::Symbol=Symbol(get(ENV, "RETESTITEMS_LOGS", default_log_display_mode(report, nworkers))),
```
nickrobinson251, line 206: Can you add a test please?

i think we can test this really easily by just replacing the test-specific `LOG_DISPLAY` with a check for the new `RETESTITEMS_LOGS` variable (in `test/log_capture.jl` and `test/_test_log_capture.jl`)

nickrobinson251 (approved):
Please bump the Project.toml version so we can release this straight away

thanks!

src/log_capture.jl
kpamnany: Turns out it doesn't. Duh.

Project.toml
```diff
@@ -1,6 +1,6 @@
 name = "ReTestItems"
 uuid = "817f1d60-ba6b-4fd5-9520-3cf149f6a823"
-version = "1.18.0"
+version = "1.18.1"
```
nickrobinson251, line 3:
```suggestion
version = "1.19.0"
```

PR #115, opened 2023-10-09
Disable worker retarts for mem pressure on Mac

nickrobinson251:


NHDaly (approved):
LGTM for now, yeah. Thanks nick!

Issue #97, opened 2023-07-20
Conditional tests (skipping testitems)


maleadt:
Some tests of a project of mine can only run on specific Julia versions. Currently, I have:

```julia
if LLVM.has_orc()
@testitem "orc" begin
...
end
end
```

ReTestItems.jl does not like this: ``LoadError: Test files must only include `@testitem` and `@testsetup` calls.``

Putting the conditional within the testitem doesn't always work, e.g,. when using types that are only conditionally defined:

```julia
@testitem "orc" begin
if has_orc()

function do_something(x::SomeOrcType)
...
end

end
end
```

I could make the test `@static`, but that's not really intuitive to most users. Furthermore, it still results in the skipped test item being shown on stdout, with a warning that it doesn't contain any tests:

```
22:22:26 | maxrss  1.0% | mem  4.7% | START (18/26) test item "orc" at test/orc_tests.jl:1
22:22:26 | maxrss  1.0% | mem  4.7% | DONE  (18/26) test item "orc" <0.1 secs (85.4% compile), 1.88 K allocs (113.380 KB)
┌ Warning: Test item "orc" at test/orc_tests.jl:1 contains test sets without tests:
│ "orc"
```

Is there another workaround? Maybe `@testitem` should support passing-in a `skip` boolean or so?

nickrobinson251:
yeah, this sounds like a nice feature! So far, in the couple cases something like this could have been used, i've just seen the `if ...` block put inside the `@testitem`, which as you say has the downside of noisy logs in stdout

i guess we'd want the `skip` keyword to take an expression (evaluating to a boolean) that's evaluated on the worker immediately before the testitem would have been run?

Drvi:
One workaround would be to use `@static if` instead of `if`, but I'm not sure if that works in your example.

Another workaround would be to give each test item a tag (like `:needs_orc`) and then implement the filtering logic in the `runtests` a la:
```julia
tags_to_filter_out = get_tags_to_filter_out(...)
runtests(PkgName) do ti
    isdisjoint(ti.tags, tags_to_filter_out)
end
```

nickrobinson251:
another use-case for this we just ran into, is skipping certain tests on pre-release versions on Julia e.g. `skip=!isempty(VERSION.prerelease)`

PR #117, opened 2023-11-04
Add support for skipping testitems

nickrobinson251:
- close https://github.com/JuliaTesting/ReTestItems.jl/issues/97

### DIFF
--- a/README.md
+++ b/README.md
@@ -133,7 +133,7 @@ it is not run before every `@testitem` that depends on the setup.
 #### Skipping tests
 
 The `skip` keyword can be used to skip a `@testitem`, meaning no code inside that test-item will run.
-A skipped test-item logs that it is being skipped and records a single "skipped" test result.
+A skipped test-item logs that it is being skipped and records a single "skipped" test result, similar to `@test_skip`.
 
 ```julia
 @testitem "skipped" skip=true begin
@@ -141,16 +141,15 @@ A skipped test-item logs that it is being skipped and records a single "skipped"
 end
 ```
 
-If `skip` is an given an `Expr`, it will be run in a new module similar to a test-item, and must return a `Bool`.
+If `skip` is an given an `Expr`, it must return a `Bool` indicating whether or not to skip the test-item.
+This expression will be run in a new module similar to a test-item immediately before the test-item would be run.
 
 ```julia
-# Don't run "orc v1" tests if we don't have orc v2
+# Don't run "orc v1" tests if we don't have orc v1
 @testitem "orc v1" skip=:(using LLVM; !LLVM.has_orc_v1()) begin
     # tests
 end
 ```
-
-Skipped test-items appear in the test results as a single skipped test, similar to `@test_skip`.
 
 #### Post-testitem hook
 
--- a/README.md
+++ b/README.md
@@ -130,6 +130,28 @@ end
 The `setup` is run once on each worker process that requires it;
 it is not run before every `@testitem` that depends on the setup.
 
+#### Skipping tests
+
+The `skip` keyword can be used to skip a `@testitem`, meaning no code inside that test-item will run.
+A skipped test-item logs that it is being skipped and records a single "skipped" test result.
+
+```julia
+@testitem "skipped" skip=true begin
+    @test false
+end
+```
+
+If `skip` is an given an `Expr`, it will be run in a new module similar to a test-item, and must return a `Bool`.
+
+```julia
+# Don't run "orc v1" tests if we don't have orc v2
+@testitem "orc v1" skip=:(using LLVM; !LLVM.has_orc_v1()) begin
+    # tests
+end
+```
+
+Skipped test-items appear in the test results as a single skipped test, similar to `@test_skip`.
+
 #### Post-testitem hook
 
 If there is something that should be checked after every single `@testitem`, then it's possible to pass an expression to `runtests` using the `test_end_expr` keyword.
--- a/README.md
+++ b/README.md
@@ -60,7 +60,17 @@ julia> runtests(
        )
 ```
 
-You can use the `name` keyword, to select test-items by name.
+For interactive sessions, all logs from the tests will be printed out in the REPL by default.
+You can disable this by passing `logs=:issues` in which case logs from a test-item are only printed if that test-items errors or fails.
+`logs=:issues` is also the default for non-interactive sessions.
+
+```julia
+julia> runtests("test/Database/"; logs=:issues)
+```
+
+#### Filtering tests
+
+You can use the `name` keyword to select test-items by name.
 Pass a string to select a test-item by its exact name,
 or pass a regular expression (regex) to match multiple test-item names.
 
@@ -70,12 +80,19 @@ julia> runtests("test/Database/"; name="issue-123")
 julia> runtests("test/Database/"; name=r"^issue")
 ```
 
-For interactive sessions, all logs from the tests will be printed out in the REPL by default.
-You can disable this by passing `logs=:issues` in which case logs from a test-item are only printed if that test-items errors or fails.
-`logs=:issues` is also the default for non-interactive sessions.
+You can pass `tags` to select test-items by tag.
+When passing multiple tags a test-item is only run if it has all the requested tags.
 
 ```julia
-julia> runtests("test/Database/"; logs=:issues)
+# Run tests that are tagged as both `regression` and `fast`
+julia> runtests("test/Database/"; tags=[:regression, :fast])
+```
+
+Filtering by `name` and `tags` can be combined to run only test-items that match both the name and tags.
+
+```julia
+# Run tests named `issue*` which also have tag `regression`.
+julia> runtests("test/Database/"; tags=:regression, name=r"^issue")
 ```
 
 ## Writing tests
@@ -150,6 +167,10 @@ This expression will be run in a new module similar to a test-item immediately b
     # tests
 end
 ```
+
+The `skip` keyword allows you to define the condition under which a test needs to be skipped,
+for example if is only expected to run on a certain platform.
+See [filtering tests](#filtering-tests) for controlling which tests run in a particular `runtests` call.
 
 #### Post-testitem hook
 
--- a/README.md
+++ b/README.md
@@ -141,7 +141,7 @@ A skipped test-item logs that it is being skipped and records a single "skipped"
 end
 ```
 
-If `skip` is an given an `Expr`, it must return a `Bool` indicating whether or not to skip the test-item.
+If `skip` is given as an `Expr`, it must return a `Bool` indicating whether or not to skip the test-item.
 This expression will be run in a new module similar to a test-item immediately before the test-item would be run.
 
 ```julia
--- a/README.md
+++ b/README.md
@@ -169,7 +169,7 @@ end
 ```
 
 The `skip` keyword allows you to define the condition under which a test needs to be skipped,
-for example if is only expected to run on a certain platform.
+for example if it can only be run on a certain platform.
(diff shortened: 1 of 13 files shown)


README.md
```diff
@@ -130,6 +130,27 @@ end
...
+A skipped test-item logs that it is being skipped and records a single "skipped" test result, similar to `@test_skip`.
+
+```julia
+@testitem "skipped" skip=true begin
+    @test false
+end
+```
+
+If `skip` is an given an `Expr`, it must return a `Bool` indicating whether or not to skip the test-item.
+This expression will be run in a new module similar to a test-item immediately before the test-item would be run.
```
NHDaly, line 145:
```suggestion
If `skip` is given as an `Expr`, it must return a `Bool` indicating whether or not to skip the test-item.
This expression will be run in a new module similar to a test-item immediately before the test-item would be run.
```

NHDaly: technically I guess this should be ```must evaluate to a `Bool` ```, rather than `return`, but I think maybe the way you have it is easier to understand?

README.md
```diff
@@ -130,6 +130,27 @@ end
...
+
+```julia
+# Don't run "orc v1" tests if we don't have orc v1
+@testitem "orc v1" skip=:(using LLVM; !LLVM.has_orc_v1()) begin
+    # tests
+end
+```
```
NHDaly, line 152: It still seems like this could just be `tag=:orc_v1`, and then in the runtests invocation you write
```julia
if LLVM.has_orc_v1()
    runtests(label=:orc_v1)
end
```
?

But I guess maybe it's nicer to have the skip logic at the testitem's definition?

src/macros.jl
```diff
@@ -120,6 +120,7 @@ struct TestItem
     setups::Vector{Symbol}
     retries::Int
     timeout::Union{Int,Nothing} # in seconds
+    skip::Union{Bool,Expr}
```
NHDaly, line 123: Should this just be `skip::Any`?

e.g. what if it's `skip=SKIP_ORC_1` and then that's defined as a global variable before the runtests? (in which case the type would be `Symbol`.) Or maybe that wouldn't work anyway since we run this in a custom module...?

How about `skip=1`, which technically
```julia
julia> 1 == true
true
```

src/ReTestItems.jl
```diff
@@ -849,6 +849,40 @@ end
...
+# Log that we skipped the testitem, and record a "skipped" test result with empty stats.
+function skiptestitem(ti::TestItem, ctx::TestContext; verbose_results::Bool=true)
+    ts = DefaultTestSet(ti.name; verbose=verbose_results)
+    Test.record(ts, Test.Broken(:skipped, ti.name))
+    push!(ti.testsets, ts)
+    stats = PerfStats()
+    push!(ti.stats, stats)
+    log_testitem_skipped(ti, ctx.ntestitems)
+    return TestItemResult(ts, stats)
+end
```
NHDaly, line 883: Should this be `!` since we're modifying `.testsets` and `.stats`? I always struggle with the border cases of this convention.

src/macros.jl
```diff
@@ -228,13 +229,26 @@ Note that `timeout` currently only works when tests are run with multiple worker
...
     end
+
+If a `@testitem` needs to be skipped, then you can set the `skip` keyword.
+Either pass `skip=true` to unconditionally skip the test item, or pass `skip` an
+expression that returns a `Bool` to determine if the testitem should be skipped.
+
+    @testitem "Skip on old Julia" skip=VERSION < v"1.9" begin
+        v = [1]
+        @test 0 == @allocations sum(v)
+    end
```
NHDaly, line 240: Wow, i can see how precedence makes that work, but i found this whitespace very jarring. Would you accept parentheses here?
```suggestion
    @testitem "Skip on old Julia" skip=(VERSION < v"1.9") begin
        v = [1]
        @test 0 == @allocations sum(v)
    end
```

NHDaly (approved):

Drvi (commented):
Just a couple of observations / thoughts. 

I think an interesting design aspect here is that a new result state is introduced: `skipped`. Unlike "manual skipping" via `shouldrun` this:
a) prints a log per skipped testitem
b) adds a broken test to the final summary
c) decides whether to run the test item on worker

I think these are very reasonable for the CI use case, for sure.

It makes me think whether we intend `skip` to be used interactively, as there might be some annoyances with verbosity of the `SKIP` logs and managing what is available to the scope where decide whether to `skip`. There are 4 ways I can think of the user can reference things for `skip`:
a) using literals / constants
b) using modules available from the test environment
c) `ENV` (which we propagate from the coordinator, IIRC)
d) referencing globals from `setup_worker_expr` (I think globals defined here would be visible?)

If we don't intend this to be run interactively, we probably want to make sure that `skip`ping on workers and `shouldrun` on the coordinator don't interact counterintuitively or maybe document the precedence?

nickrobinson251:
The design here is for `skip` to act similarly to `@test_skip`, which is why skipped testitems log that they are skipped.

i'm not sure i fully understand the concern about interactive use-case and `shouldrun`. Can you give an example?

Here's what it looks like on a few cases at the REPL
```julia
julia> runtests("test/testfiles/_skip_tests.jl"; name=r"skip true")
[ Info: Scanning for test items in project `ReTestItems` at paths: test/testfiles/_skip_tests.jl
[ Info: Finished scanning for test items in 0.03 seconds. Scheduling 1 tests on pid 57695
11:32:59 | SKIPPED (1/1) test item "skip true" at test/testfiles/_skip_tests.jl:11
Test Summary:                       | Broken  Total  Time
ReTestItems                         |      1      1  0.1s
  test                              |      1      1
    test/testfiles                  |      1      1
      test/testfiles/_skip_tests.jl |      1      1
        skip true                   |      1      1

julia> runtests("test/testfiles/_skip_tests.jl"; name=r"no skip")
[ Info: Scanning for test items in project `ReTestItems` at paths: test/testfiles/_skip_tests.jl
[ Info: Finished scanning for test items in 0.04 seconds. Scheduling 1 tests on pid 57695
11:33:02 | START (1/1) test item "no skip, 1 pass" at test/testfiles/_skip_tests.jl:2
11:33:02 | DONE (1/1) test item "no skip, 1 pass" <0.1 secs (<0.1% compile), 1.57 K allocs (114.360 KB)
Test Summary:                       | Pass  Total  Time
ReTestItems                         |    1      1  0.1s
  test                              |    1      1
    test/testfiles                  |    1      1
      test/testfiles/_skip_tests.jl |    1      1
        no skip, 1 pass             |    1      1  0.0s

julia> runtests("test/testfiles/_skip_tests.jl"; name=r"skip expr", nworkers=2)
[ Info: Scanning for test items in project `ReTestItems` at paths: test/testfiles/_skip_tests.jl
[ Info: Finished scanning for test items in 0.04 seconds. Scheduling 3 tests on pid 57695 with 2 worker processes and 2 threads per worker.
[ Info: Starting test workers
  Worker 60932:  [ Info: Starting test worker 1 on pid = 60932, with 2 threads
  Worker 60931:  [ Info: Starting test worker 2 on pid = 60931, with 2 threads
[ Info: Starting running test items
  Worker 60931:  11:33:49 | SKIPPED (2/3) test item "skip expr true" at test/testfiles/_skip_tests.jl:21
  Worker 60932:  11:33:49 | START (1/3) test item "skip expr false, 2 pass" at test/testfiles/_skip_tests.jl:16
  Worker 60932:  11:33:49 | DONE (1/3) test item "skip expr false, 2 pass" <0.1 secs (91.9% compile), 45.88 K allocs (3.242 MB)
  Worker 60931:  11:33:49 | SKIPPED (3/3) test item "skip expr block true" at test/testfiles/_skip_tests.jl:25
  Worker 60931:  [60931] signal (15): Terminated: 15
in expression starting at none:1

Test Summary:                       | Pass  Broken  Total  Time
ReTestItems                         |    2       2      4  2.8s
  test                              |    2       2      4
    test/testfiles                  |    2       2      4
      test/testfiles/_skip_tests.jl |    2       2      4
        skip expr false, 2 pass     |    2              2  0.1s
        skip expr true              |            1      1
        skip expr block true        |            1      1
```

src/log_capture.jl
```diff
@@ -241,35 +241,44 @@ function _print_test_errors(report_iob, ts::DefaultTestSet, worker_info)
...
 
-# Marks the start of each test item
-function log_testitem_start(ti::TestItem, ntestitems=0)
-    io = IOContext(IOBuffer(), :color => get(DEFAULT_STDOUT[], :color, false)::Bool)
+function print_state(io, state, ti, ntestitems)
     interactive = parse(Bool, get(ENV, "RETESTITEMS_INTERACTIVE", string(Base.isinteractive())))
     print(io, format(now(), "HH:MM:SS | "))
     !interactive && print(io, _mem_watermark())
-    printstyled(io, "START"; bold=true)
+    printstyled(io, uppercase(state); bold=true)
```
Drvi, line 248: This refactor seems to break the alignment we had before. I.e.:
```
  Worker 60931:  11:33:49 | SKIPPED (2/3) test item "skip expr true" at test/testfiles/_skip_tests.jl:21
  Worker 60932:  11:33:49 | START (1/3) test item "skip expr false, 2 pass" at test/testfiles/_skip_tests.jl:16
  Worker 60932:  11:33:49 | DONE (1/3) test item "skip expr false, 2 pass" <0.1 secs (91.9% compile), 45.88 K allocs (3.242 MB)
```
I think it would be preferable to have:
```
  Worker 60931:  11:33:49 | SKIP  (2/3) test item "skip expr true" at test/testfiles/_skip_tests.jl:21
  Worker 60932:  11:33:49 | START (1/3) test item "skip expr false, 2 pass" at test/testfiles/_skip_tests.jl:16
  Worker 60932:  11:33:49 | DONE  (1/3) test item "skip expr false, 2 pass" <0.1 secs (91.9% compile), 45.88 K allocs (3.242 MB)
```

Drvi:
Example for spammy logging: if there are 2000 test items with skip conditions that are affected by an ENV variable and the user sets the ENV to only run 3 test items, they'd get 1997 lines printed in their terminal they don't really care about (`shouldrun` filtering would not produce the 1997 `SKIPPED` logs). Maybe the solution is just not to use the skipping to interactively subset the number of tests to run.

For the `skip` vs `shouldrun` -- we give users two mechanism to dynamically select which tests to actually evaluate and I just want to make sure they don't lead to unexpected interactions. E.g. if the user specifies `skip` only run when something very overly specific to the CI is true, then there is no easy way to override that beside trying to emulate that specific CI thing locally or changing the `skip` condition; which made me think that e.g. when the user singles out a test item like:
```julia
untests("test"; name="test that skips")
```
But I think your point about this being very similar to `@test_skip` is a good one and the user should just make the skipping condition flexible enough. :+1:

README.md
nickrobinson251: yeah, talk to `evaluate` has been griped about https://github.com/JuliaTesting/ReTestItems.jl/commit/0b7275325db58eba956e1d23a375a81fc34a8efe 😆

src/macros.jl
nickrobinson251: i'm okay excluding `skip=1` and requiring a `Bool`. In general Julia requires a `Bool`for boolean things (like conditionals).

src/ReTestItems.jl
nickrobinson251: i decided against it to match with `runtestitem` (which is unfortunately an exported name from a time before individual test-items auto-run in the repl)

i think it's just about acceptable because it modifies "private" fields (in contrast to "public" fields like `tags`, `name`, etc)

src/log_capture.jl
nickrobinson251: thanks, fixed with a test

PR #119, opened 2023-11-11
Add v1.10 to CI

nickrobinson251:


.github/workflows/CI.yml
```diff
@@ -19,6 +19,7 @@ jobs:
         version:
           - '1.8'
           - '1.9'
+          - '1.10-nightly'
```
NHDaly, line 22: Should we also do `'nightly'`? (i.e. julia main?)

NHDaly (approved):

.github/workflows/CI.yml
nickrobinson251: we already do, in a separate CI job so it doesn't turn the README "CI" badge red: https://github.com/JuliaTesting/ReTestItems.jl/blob/main/.github/workflows/JuliaNightly.yml

PR #129, opened 2023-12-07
Allow `@testitem` to set its own timeout

nickrobinson251:
- close https://github.com/JuliaTesting/ReTestItems.jl/issues/13
- an `@testitem` can set its own timeout (in seconds) via the `timeout` keyword, like `@testitem "foo" timeout=600`
- this value takes precedence over the `testitem_timeout` passed to `runtests(...)` (i.e. this PR makes `testitem_timeout` be a "default" timeout value that can be overridden per testitem.)
  - The alternative would be to take the `max` of the two values, which is what we do for `retries` (so that inconsistency is perhaps a little unfortunate).
  - But giving the `@testitem`'s own `timeout=...` value precedence allows a `@testitem` to set a _lower_ limit than the default, which is useful for tests that should go fast but are at risk of blowup.

### DIFF
--- a/src/ReTestItems.jl
+++ b/src/ReTestItems.jl
@@ -490,8 +490,6 @@ function manage_worker(
             worker = robust_start_worker(proj_name, nworker_threads, worker_init_expr, ntestitems)
         end
         testitem.workerid[] = worker.pid
-        # TODO: should the testitem's own timeout take precedence or should we just take the `max`?
-        # timeout = max(something(testitem.timeout, 0.0), default_timeout)
         timeout = something(testitem.timeout, default_timeout)
         fut = remote_eval(worker, :(ReTestItems.runtestitem($testitem, GLOBAL_TEST_CONTEXT; test_end_expr=$(QuoteNode(test_end_expr)), verbose_results=$verbose_results, logs=$(QuoteNode(logs)))))
         max_runs = 1 + max(retries, testitem.retries)
--- a/src/ReTestItems.jl
+++ b/src/ReTestItems.jl
@@ -444,7 +444,7 @@ function record_timeout!(testitem, run_number::Int, timeout_s::Int)
         end
     end
     msg = "Timed out after $time_str running test item $(repr(testitem.name)) (run=$run_number)"
-    record_test_error!(testitem, msg, timeout_limit)
+    record_test_error!(testitem, msg, timeout_s)
 end
 
 function record_worker_terminated!(testitem, worker::Worker, run_number::Int)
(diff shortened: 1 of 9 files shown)


src/ReTestItems.jl
```diff
@@ -487,6 +490,9 @@ function manage_worker(
             worker = robust_start_worker(proj_name, nworker_threads, worker_init_expr, ntestitems)
         end
         testitem.workerid[] = worker.pid
+        # TODO: should the testitem's own timeout take precedence or should we just take the `max`?
+        # timeout = max(something(testitem.timeout, 0.0), default_timeout)
+        timeout = something(testitem.timeout, default_timeout)
```
mcmcgrath13, line 495: user should definitely take precedence. It can be useful to set a lower bound here so that for tests that should go fast but are at risk of blowup, you can set a lower timeout than the default so those timeouts can catch regressions more often

src/macros.jl
```diff
@@ -219,10 +220,19 @@ If a `@testitem` passes on retry, then it will be recorded as passing in the tes
     @testitem "Flaky test" retries=1 begin
         @test rand() < 1e-4
     end
+
+If a `@testitem` should be aborted after a certain period of time, e.g. the test is known
+to occassionally hang, then you can set a timeout (in seconds) by passing the `timeout` keyword.
+Note that `timeout` currently only works when tests are run with multiple workers.
```
mcmcgrath13, line 226: why is that? (out of curiosity)

test/testfiles/_timeout2_tests.jl
```diff
@@ -0,0 +1,4 @@
+@testitem "Sets timeout=6" timeout=6 begin
```
mcmcgrath13, line 1: I am probably missing something, but where is this used? it should fail, right?

mcmcgrath13 (approved):
thank you!

src/ReTestItems.jl
```diff
@@ -428,7 +431,7 @@ end
 
 any_non_pass(ts::DefaultTestSet) = ts.anynonpass
 
-function record_timeout!(testitem, run_number::Int, timeout_limit::Real)
+function record_timeout!(testitem, run_number::Int, timeout_limit::Float64)
     timeout_s = round(Int, timeout_limit)
```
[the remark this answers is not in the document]
nickrobinson251: yah, this is a hassle. it's just because we were lax before and so were allowing users to pass any `Real`, so in this PR i converted them to Float64 asap, so we at least know the type we have... but i think going the other way and rounding them to Int is the better idea (we still need to accept Floats to avoid a breaking change, as you say)

src/macros.jl
Drvi: I think to robustly cancel a task on the same process, we'd need something like https://github.com/JuliaLang/julia/issues/6283

PR #130, opened 2023-12-08
[RFC] Allow `timeout` to take an expression like `60*2`

nickrobinson251:
Because
```julia
@testitem "TI" timeout=60*15 begin
    ...
end
```
is a lot clearer than
```julia
@testitem "TI" timeout=900 begin
    ...
end
```

### DIFF
--- a/src/macros.jl
+++ b/src/macros.jl
@@ -146,7 +146,7 @@ function TestItem(number, name, id, tags, default_imports, setups, retries, time
 end
 
 """
-    @testitem "name" [tags=[] setup=[] retries=0 skip=false default_imports=true] begin
+    @testitem "name" [tags=[] setup=[] retries=0 skip=false timeout=nothing default_imports=true] begin
         # code that will be run as tests
     end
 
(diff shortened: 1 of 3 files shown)


src/macros.jl
```diff
@@ -146,7 +146,7 @@ function TestItem(number, name, id, tags, default_imports, setups, retries, time
 end
 
 """
-    @testitem "name" [tags=[] setup=[] retries=0 skip=false default_imports=true] begin
+    @testitem "name" [tags=[] setup=[] retries=0 skip=false timeout=nothing default_imports=true] begin
```
Drvi, line 149: This got me thinking that `nothing` is a valid value the user can provide, which I don't think is the case. Maybe `timeout=0` could mean no timeout?

Drvi (approved):
LGTM

src/macros.jl
nickrobinson251: `timout=nothing` is a valid input

nickrobinson251: indeed, it is the default value, and means this testitem has no specific timeout (so inherits the global timeout)

Currently the way to disable timeouts all together is to just pass an arbitrarily large value for the timeout, i.e. `testitem_timeout=Inf` (but that's unchanged by this PR)

Drvi: I mean as valid input representing the default case for "no timeout".
```julia
julia> using ReTestItems

julia> @testitem "aaa" timeout=nothing begin end
ERROR: LoadError: AssertionError: `timeout` keyword must be passed a `Real`, got `timeout=nothing`
```

nickrobinson251: ahh, i see

Drvi: `Inf` is also a good choice, but it currently won't work because it will reach this codepath not as a `Real`, nor an `Expr` but as a `Symbol` (I just checked it as I didn't understand the error I was getting) and then we are checking for `timeout = ceil(Int, t)` which also doesn't play nicely with an `Inf`.

nickrobinson251: y'know, the more i think about this PR, the less i like it

I really don't like arbitrary user-given code being able to run at "testitem definition" time, before we start workers and running things in parallel

I think i might park this PR for now... `30*60` is a lot clearer than `1800`, but idk if it's worth the cost

thanks for poking at this PR!

NHDaly: 🤔 could you evaluate the expression in the context of the worker instead? Forward the expression along to the worker and then evaluate it there as the first thing it does?

I'm not sure it helps all that much, but it addresses part of the concern you raised in your last comment.

But yeah parking this seems fine too

PR #131, opened 2023-12-14
Add option to throw rather than warn if given a useless path

nickrobinson251:
If `runtests` is given a path that does not exist or a path to a file that's not got one of the valid suffixes, then these paths cannot possibly lead to any test-items or test-setups being loaded.

Current behaviour is to produce a warning about these paths, and then continue (in case we have been given other paths that do lead to tests).

But we have found this behaviour in practice leads to an accumulation of dead paths over time, as files or directories get removed or renamed. (cc @charnik)

So this PR adds the ability to opt-in to throwing an error, rather than just logging a warning, if `runtests` is passed any path that cannot possibly result in tests.

For this not be a breaking change, we gate the new behaviour behind the `validate_paths` keyword, which can also be set via the `RETESTITEMS_VALIDATE_PATHS` env variable.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "ReTestItems"
 uuid = "817f1d60-ba6b-4fd5-9520-3cf149f6a823"
-version = "1.21.0"
+version = "1.22.0"
 
 [deps]
 Dates = "ade2ca70-3891-5945-98fb-dc099432e06a"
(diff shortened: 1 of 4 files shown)


Drvi (approved):
Good idea 👍

NHDaly:
Good idea 👍 `go test` has the same behavior: it throws an error and aborts if one of the paths don't exist (red means returned non-0):

Screenshot 2023-12-14 at 9 58 17 AM

PR #149, opened 2024-03-12
Try not using gensym for testsetup module names

nickrobinson251:
current using `@testsetup module` means error messages/tracebacks that refer to it get really ugly names like 
`Main.var"##FDSetup#298"`, eg:
```jl
    [2] fd_test(::Type, ::Vararg{Any}; verbose::Bool)
      @ Main.var"##FDSetup#298" ~/base-fds/test/base-fd-tests.jl:259
```


rntz:
Any idea what went wrong with the tests here

rntz:
looks like the tests that failed were precisely the ones testing log capture?

PR #150, opened 2024-03-18
Defer setting up the temp folder for pkgimage relocability

Drvi:


nickrobinson251 (approved):

src/workers.jl
```diff
@@ -183,6 +183,7 @@ function Worker(;
         if isassigned(GLOBAL_CALLBACK_PER_WORKER)
             GLOBAL_CALLBACK_PER_WORKER[](w)
         end
+        remote_fetch(w, :(1+1)) # Wait for worker to be ready
```
nickrobinson251, line 186: thanks for fixing this! i wonder if there's a more elegant way to know the worker is ready?

nickrobinson251 (approved):
LGTM

src/workers.jl
Drvi: I think the canonical way would be to let the tasks tell you that they are ready (e.g. they could notify a condition and decrement a counter, or we could use multiple Events, or they put something into a channel...). I think trying the three Events solution won't be too bad, I'll try it:+1:

PR #155, opened 2024-04-19
Minor fixup to Gentle Close PR: Wait until the worker closes itself, before force-closing it in response to the closed pipes.

NHDaly:
When we request a worker to gracefully close, the worker closes its pipes and then shutsdown.

Without this small commit, the coordinator would detect the closed pipes, and call terminate! in response.

This change tells the coordinator: "Hey, don't worry, _we asked_ it to shut down! Just chill out, and wait for it to go to sleep peacefully, my friend."

BUT due to https://github.com/JuliaLang/julia/issues/54145, we have to introduce a less-reliable `sleep(5)` on macos. :/
That should be long enough to cover it though, and even if this is too slow, all that will happen is the worker will die with SIG15 instead of exiting 0.

### DIFF
--- a/src/workers.jl
+++ b/src/workers.jl
@@ -59,6 +59,7 @@ mutable struct Worker
     process_watch::Task
     futures::Dict{UInt64, Future} # Request.id -> Future
     @atomic terminated::Bool
+    gracefully_closed::Bool
 end
 
 # used to close Future.value channels when a worker terminates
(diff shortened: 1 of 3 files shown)


src/workers.jl
```diff
@@ -81,6 +82,16 @@ function terminate!(w::Worker, from::Symbol=:manual)
         end
         empty!(w.futures)
     end
+    # Give some time for the worker to exit gracefully.
+    if w.gracefully_closed
+        @static if !Sys.isapple()
+            # NOTE: THIS CAUSES A HANG ON MACOS! We can remove this if-check once
+            # the julia issue is resolved:
```
quinnj, line 89: link to Julia issue #?

quinnj (approved):
LGTM; it would be great to sync these changes to ConcurrentUtilities.jl as well!

quinnj:
Hmmm, macos 1.9/1.10 failures are both on the `test retrying failing testitem` tests, which do involve having workers die and re-trying the tests, so possibly related? Probably worth trying to reproduce locally to confirm.

Issue #105, opened 2023-08-07
Print profile dump before killing worker when we timeout


nickrobinson251:
When a testitem hits the timeout limit, we print out whatever logs it produced and then kill the worker, but we don't print out any info about what the worker was actually doing, which makes it hard to know e.g. if the testitem was deadlocked.

It would be helpful to collect and print out a CPU profile [by sending a SIGINFO/SIGUSR1](https://github.com/JuliaLang/julia/blob/5544a0fab7648cfa61fe79cd557a7504a92ec1b5/NEWS.md?plain=1#L185-L187) to the process before sending the SIGTERM

Alternatively, we could dump julia task backtraces and/or CPU stacktraces.

PR #157, opened 2024-05-06
Get thread and task backtraces before terminating a worker on timeout

kpamnany:
Implements the alternative described in https://github.com/JuliaTesting/ReTestItems.jl/issues/105 as an option.

> Alternatively, we could dump julia task backtraces and/or CPU stacktraces.

I think a similar incantation can be assembled for `lldb`, to support Mac. We could support FreeBSD/OpenBSD since they have `gdb`, but I don't have a machine that would let me test. Note that `Sys.isbsd()` returns `true` for Mac.

### DIFF
--- a/src/ReTestItems.jl
+++ b/src/ReTestItems.jl
@@ -572,14 +572,14 @@ function manage_worker(
             if e isa TimeoutException
                 @warn "$worker timed out running test item $(repr(testitem.name)) after $timeout seconds."
                 if timeout_profile_wait > 0
-                    @warn "Gathering a CPU profile from the worker."
+                    @info "Gathering a CPU profile from $worker."
                     trigger_profile(worker, timeout_profile_wait, :timeout)
                 end
                 if timeout_backtraces
-                    @warn "Gathering thread and task backtraces from the worker."
+                    @info "Gathering thread and task backtraces from $worker."
                     trigger_backtraces(worker, :timeout)
                 end
-                @warn "Terminating the worker."
+                @info "Terminating $worker."
                 terminate!(worker, :timeout)
                 wait(worker)
                 # TODO: We print the captured logs after the worker is terminated,
--- a/src/ReTestItems.jl
+++ b/src/ReTestItems.jl
@@ -197,6 +197,9 @@ will be run.
    Zero means no profile will be taken. Can also be set using the `RETESTITEMS_TIMEOUT_PROFILE_WAIT`
    environment variable. See the [Profile documentation](https://docs.julialang.org/en/v1/stdlib/Profile/#Triggered-During-Execution)
    for more information on triggered profiles. Note you can use `worker_init_expr` to tweak the profile settings on workers.
+- `timeout_backtraces::Bool=false`: Linux only. If `true`, we use GDB to get thread and task
+  backtraces from a worker that times-out. Can also be set using the `RETESTITEMS_TIMEOUT_BACKTRACES`
+  environment variable. If a CPU profile was also requested, this runs after that is taken.
 """
 function runtests end
 
@@ -243,6 +246,7 @@ function runtests(
     test_end_expr::Expr=Expr(:block),
     validate_paths::Bool=parse(Bool, get(ENV, "RETESTITEMS_VALIDATE_PATHS", "false")),
     timeout_profile_wait::Real=parse(Int, get(ENV, "RETESTITEMS_TIMEOUT_PROFILE_WAIT", "0")),
+    timeout_backtraces::Bool=parse(Bool, get(ENV, "RETESTITEMS_TIMEOUT_BACKTRACES", "false")),
 )
     nworker_threads = _validated_nworker_threads(nworker_threads)
     paths′ = _validated_paths(paths, validate_paths)
@@ -262,13 +266,14 @@ function runtests(
     timeout = ceil(Int, testitem_timeout)
     timeout_profile_wait = ceil(Int, timeout_profile_wait)
     (timeout_profile_wait > 0 && Sys.iswindows()) && @warn "CPU profiles on timeout is not supported on Windows, ignoring `timeout_profile_wait`"
+    (timeout_backtraces && !Sys.islinux()) && @warn "Backtraces on timeout is only supported on Linux, ignoring `timeout_backtraces`"
     debuglvl = Int(debug)
     if debuglvl > 0
         LoggingExtras.withlevel(LoggingExtras.Debug; verbosity=debuglvl) do
-            _runtests(shouldrun_combined, paths′, nworkers, nworker_threads, worker_init_expr, test_end_expr, timeout, retries, memory_threshold, verbose_results, debuglvl, report, logs, timeout_profile_wait)
+            _runtests(shouldrun_combined, paths′, nworkers, nworker_threads, worker_init_expr, test_end_expr, timeout, retries, memory_threshold, verbose_results, debuglvl, report, logs, timeout_profile_wait, timeout_backtraces)
         end
     else
-        return _runtests(shouldrun_combined, paths′, nworkers, nworker_threads, worker_init_expr, test_end_expr, timeout, retries, memory_threshold, verbose_results, debuglvl, report, logs, timeout_profile_wait)
+        return _runtests(shouldrun_combined, paths′, nworkers, nworker_threads, worker_init_expr, test_end_expr, timeout, retries, memory_threshold, verbose_results, debuglvl, report, logs, timeout_profile_wait, timeout_backtraces)
     end
 end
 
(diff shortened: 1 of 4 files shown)


kpamnany:
CI doesn't seem to have `gdb` so the tests aren't running. They pass locally, FWIW.

src/ReTestItems.jl
```diff
@@ -563,11 +570,16 @@ function manage_worker(
             @debugv 2 "Error" exception=e
             # Handle the exception
             if e isa TimeoutException
+                @warn "$worker timed out running test item $(repr(testitem.name)) after $timeout seconds."
                 if timeout_profile_wait > 0
-                    @warn "$worker timed out running test item $(repr(testitem.name)) after $timeout seconds. \
-                        A CPU profile will be triggered on the worker and then it will be terminated."
+                    @warn "Gathering a CPU profile from the worker."
```
nickrobinson251, line 575: Now that we're logging multiple messages, they might be interspersed with logs coming from other workers, so i think we should always interpolate the `$worker` value so we can identify which worker is being referred to

Also i think the messages after the initial "timeout" log, can be `info` since they're "normal" behaviour for the situation, and it can be nice to keep the number of `warn` messages equal to the number of things that have gone wrong (e.g. timeouts)

```suggestion
                    @info "Gathering a CPU profile from $worker."
```

src/ReTestItems.jl
```diff
@@ -563,11 +570,16 @@ function manage_worker(
...
                     trigger_profile(worker, timeout_profile_wait, :timeout)
                 end
+                if timeout_backtraces
+                    @warn "Gathering thread and task backtraces from the worker."
```
nickrobinson251, line 579:
```suggestion
                    @info "Gathering thread and task backtraces from $worker."
```

src/ReTestItems.jl
```diff
@@ -563,11 +570,16 @@ function manage_worker(
...
+                    trigger_backtraces(worker, :timeout)
+                end
+                @warn "Terminating the worker."
```
nickrobinson251, line 582:
```suggestion
                @info "Terminating $worker."
```

src/workers.jl
```diff
@@ -120,7 +120,23 @@ function trigger_profile(w::Worker, timeout_profile_wait, from::Symbol=:manual)
...
+# to get task backtraces. We wait for GDB to finish so no sleep is needed. Thread backtraces
+# are written to `gdb.btall`, which is overwritten if present.
+function trigger_backtraces(w::Worker, from::Symbol=:manual)
+    if Sys.islinux()
+        @debug "using GDB to get thread and task backtraces on worker $(w.pid) from $from"
+        gdb_cmd = `gdb -ex "handle SIGSEGV noprint nostop pass" -ex "handle SIGUSR2 noprint nostop pass" -ex "set pagination 0" -ex "set logging overwrite on" -ex "set logging file gdb.btall" -ex "set logging redirect on" -ex "set logging enabled on" -ex "thread apply all bt" -ex "set logging enabled off" -ex "call jl_print_task_backtraces(1)" --batch -p $(w.pid)`
+        try
+            run(gdb_cmd; wait = true) # if `wait = false`, std* are redirected to devnull
+        catch e
+            @warn "Could not get thread/task backtraces via GDB.\n$(sprint(showerror, e))"
```
nickrobinson251, line 138: i think the `exception=...` keyword does the right thing (i.e. calls `showerror`) -- is there a reason to prefer interpolating into the string directly?

```suggestion
            @warn "Could not get thread/task backtraces via GDB." exception=e
```

src/workers.jl
```diff
@@ -120,7 +120,23 @@ function trigger_profile(w::Worker, timeout_profile_wait, from::Symbol=:manual)
...
+        sleep(timeout_profile_wait) # Leave time to output the profile.
+    end
+    return nothing
+end
+
+# Spawn GDB to request thread backtraces followed by invoking jl_print_task_backtraces
+# to get task backtraces. We wait for GDB to finish so no sleep is needed. Thread backtraces
+# are written to `gdb.btall`, which is overwritten if present.
+function trigger_backtraces(w::Worker, from::Symbol=:manual)
+    if Sys.islinux()
```
nickrobinson251, line 132: Which part of this is Linux specific? 
- On Mac could we have an `lldb` call that achieves the same thing as `gdb`? 
- does BSD support `gdb`, so this could be `islinux() || isbsd()`?

Can you add the answer to the PR description please, so if this feature is later requested for other OS we can look back at this PR?

src/workers.jl
```diff
@@ -120,7 +120,23 @@ function trigger_profile(w::Worker, timeout_profile_wait, from::Symbol=:manual)
...
+        @debug "using GDB to get thread and task backtraces on worker $(w.pid) from $from"
+        gdb_cmd = `gdb -ex "handle SIGSEGV noprint nostop pass" -ex "handle SIGUSR2 noprint nostop pass" -ex "set pagination 0" -ex "set logging overwrite on" -ex "set logging file gdb.btall" -ex "set logging redirect on" -ex "set logging enabled on" -ex "thread apply all bt" -ex "set logging enabled off" -ex "call jl_print_task_backtraces(1)" --batch -p $(w.pid)`
```
nickrobinson251, line 134: there is a lot going on here... i think we need a comment (maybe a long one) 

perhaps we write this out over multiple lines so we can add comments about blocks of `-ex` commands, like

```suggestion
        gdb_cmd = Cmd([
        "gdb",
        # comment here about SIGSEGV and SIGUSR2
        "-ex", "handle SIGSEGV noprint nostop pass",
        "-ex", "handle SIGUSR2 noprint nostop pass",
        "-ex", "set pagination 0",
        "-ex", "set logging overwrite on",
        "-ex", "set logging file gdb.btall",
        "-ex", "set logging redirect on",
        # comment here about getting traces for all threads
        "-ex", "set logging enabled on",
        "-ex", "thread apply all bt",
        "-ex", "set logging enabled off",
        # comment here...
        "-ex", "call jl_print_task_backtraces(1)",
        "--batch", "-p", "$(w.pid)"
    ])
```

e.g. i'm thinking a block like this needs explaining (probably not to a `gdb` connoisseur but to some of us less familiar with this wizardry):
```
-ex "set logging enabled on" -ex "thread apply all bt" -ex "set logging enabled off"
```

test/integrationtests.jl
```diff
@@ -890,6 +890,68 @@ end
...
+    end
+
+    @testset "timeout_backtraces=true means we gather backtraces" begin
+    capture_timeout_backtraces(true) do logs
+        if gdb_available()
+            @test occursin("in signal_listener", logs)
+            @test count(r"pthread_cond_wait|__psych_cvwait", logs) > 0 # the stacktrace was printed (will fail on Windows)
+            @test occursin("==== Thread 1 created", logs)
+            @test occursin("==== End thread 1", logs)
+        end
```
nickrobinson251, line 939:
```suggestion
        else
            @test_skip "gdb unavailable"
        end
```

test/integrationtests.jl
```diff
@@ -890,6 +890,68 @@ end
     end
 end
 
+@testset "Backtraces timeout trigger" begin
+    function gdb_available()
+        try
+            run(`gdb -ex "exit"`)
```
nickrobinson251, line 896: is there a way to run CI on machines with `gdb` available? why don't the linux CI machines have it?

nickrobinson251 (commented):
> CI doesn't seem to have gdb so the tests aren't running

I think this is worth investigating a little before we merge this

also left a few minor suggestions

src/workers.jl
nickrobinson251: > We could support FreeBSD/OpenBSD since they have gdb, but I don't have a machine that would let me test. Note that Sys.isbsd() returns true for Mac.

There is `Sys.isopenbsd()` which is what we'd want. But i take your point about not being able to test it 👍

nickrobinson251: opened https://github.com/JuliaTesting/ReTestItems.jl/issues/158

test/integrationtests.jl
nickrobinson251: yeah, i think that just needs an `if:` line like
```
- name: Install GDB
  if: matrix.os == 'ubuntu-latest'
  run: sudo apt-get install -y gdb
```
not certain on the syntax, might need to be
```
- name: Install GDB
  if: ${{ matrix.os }} == 'ubuntu-latest'
  run: sudo apt-get install -y gdb
```

kpamnany: I added it, but the CI machine does not allow `gdb` to attach to the process:
```
Could not attach to process.  If your uid matches the uid of the target
process, check the setting of /proc/sys/kernel/yama/ptrace_scope, or try
again as the root user.  For more details, see /etc/sysctl.d/10-ptrace.conf
ptrace: Inappropriate ioctl for device.
```
So I've backed it out. 🤷‍♂️

kpamnany:
After much testing, it turns out that `gdb` is actually pretty terrible when the number of threads in the process is large. And maybe also when you run multiple `gdb`s at once?

In my latest (still running) test:
```
  PID USER      PR  NI    VIRT    RES    SHR S  %CPU  %MEM     TIME+ COMMAND
10695 kpamnany  20   0 7165224   6.0g 208860 R 100.3  20.4  87:57.87 gdb
10712 kpamnany  20   0 7259628   6.1g 208852 R 100.3  20.7  86:42.86 gdb
    1 root      20   0    2276   1416   1412 S   0.0   0.0   0:00.10 init(Ubuntu)
 ```
Whatever the cause, using `gdb` is not practical and we'll need another way to get thread/task backtraces from Julia. I'm closing this, won't work.

PR #160, opened 2024-06-27
Move GC to `test_end_expr` and only collect when `nworkers > 1`

Drvi:


kpamnany (approved):
👍

NHDaly (approved):

nickrobinson251:
Can you help me understand this change? 

Moving the GC invocation to `test_end_expr` seems to me like it leads to the surprising behaviour where someone passing their own `test_end_expr` changes how often we invoke GC. I see some internal/private discussion (https://relationalai.atlassian.net/browse/RAI-27779) that seems to conclude that we should only trigger 1 GC pass (rather than 2) and only when there are multiple workers, but AFAIU the change in this PR is different to that conclusion, because it also moves GC to the user-controlled `test_end_expr` (and introduces a GC invocation before we schedule any tests)


cc @Drvi

nickrobinson251:
Put up an suggested alternative change here, where hopefully we can discuss if/how to tweak this change to GC behaviour: https://github.com/JuliaTesting/ReTestItems.jl/pull/169

Drvi:
GCs do not cooperate across processes, so we trigger GC before we spawn any worker processes to make sure the parent process doesn't hold up to unnecessary memory since the workers won't be able to collect it if they get under heavy memory pressure. Within a single process with no workers, forcing GC shouldn't be necessary, GC should be smart enough to know when to collect. Moving the call to a `test_end_expr` gives the user the ability to opt-out from this behavior (and use a custom memory pressure handling system if they want, like OOMGuardian) and makes the default value actually meaningful.

nickrobinson251:
thanks, those first two points make a lot of sense!

I think i have two more questions:
- previously we were doing 2 GC sweeps (a 'full' one, then a partial (young-objects-only) one), i don't know why that was the case, but i see we changed it here to do a single full sweep... why is that the best choice? i.e. why do `gc(true)` rather than `gc(false)` (i don't have a good understanding here at all... so i don't have an opinion)
- having this be user-controllable but with a sensible default sounds great to me. what's the benefit of having this be controlled via `test_end_expr` as opposed to a separate `gc_between_testitems::Bool=(nworkers > 1)` keyword? Currently, having it controlled by `test_end_expr`, i think leads to the surprising behaviour where passing in a `test_end_expr` means you know also have to think about GC (since you'd no longer get the default `gc(true)` if you have multiple workers). So, i'm leaning towards thinking it's even better to move this to its own keyword - what do you think?

Drvi:
>previously we were doing 2 GC sweeps (a 'full' one, then a partial (young-objects-only) one), i don't know why that was the case, but i see we changed it here to do a single full sweep... why is that the best choice? i.e. why do gc(true) rather than gc(false) (i don't have a good understanding here at all... so i don't have an opinion)

Basically, Diogo told us that we don't need to do two GC calls to do a full collection, we misunderstood how the GC worked earlier.

>having this be user-controllable but with a sensible default sounds great to me. what's the benefit of having this be controlled via test_end_expr as opposed to a separate gc_between_testitems::Bool=(nworkers > 1) keyword? Currently, having it controlled by test_end_expr, i think leads to the surprising behaviour where passing in a test_end_expr means you know also have to think about GC (since you'd no longer get the default gc(true) if you have multiple workers). So, i'm leaning towards thinking it's even better to move this to its own keyword - what do you think?

My hope was not to increase the API surface area and it felt like the GC call is a good candidate for the default value of `test_end_expr`. But you have the advantage of returning to this codebase after (I hope) having a break from it, so you are better suited to know what behavior is surprising and unexpected:) So I'm happy to defer this decision to you.

PR #162, opened 2024-07-12
[RAI-28133] Output an error when `worker_init_expr` is set but test workers are zero

nantiamak:
Currently, if test workers are zero, `worker_init_expr` won't run even if it's defined.
This PR returns an error in this case.

A `@testset` is added in `test/integrationtests.jl`.


NHDaly (approved):
LGTM thanks Nantia! 👍

PR #163, opened 2024-07-12
Always run the test_end_expr, even if the test throws an exception.

NHDaly:
Put the test_end_expr running in a try/finally.

Also adds a debugv log for running the test_end_expr.
Adds tests.

--------

I made the following design decision:
- We run the test_end_expr iff we run the testitem.
    - i.e. If the testsetup fails or does not exist, we do not start the testitem and therefore do not run the test_end_expr.

Let me know if that doesn't sound right to you.

### DIFF
--- a/src/ReTestItems.jl
+++ b/src/ReTestItems.jl
@@ -935,10 +935,16 @@ end
 # when `runtestitem` called directly or `@testitem` called outside of `runtests`.
 function runtestitem(
     ti::TestItem, ctx::TestContext;
-    test_end_expr::Expr=Expr(:block), logs::Symbol=:eager, verbose_results::Bool=true, finish_test::Bool=true,
+    test_end_expr::Union{Nothing,Expr}=nothing, logs::Symbol=:eager, verbose_results::Bool=true, finish_test::Bool=true,
 )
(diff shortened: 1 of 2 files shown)


Drvi (approved):
LGTM, thanks!

nickrobinson251 (approved):
Thanks! LGTM

> We run the test_end_expr iff we run the testitem.
>  i.e. If the testsetup fails or does not exist, we do not start the testitem and therefore do not run the test_end_expr.

I think that's a good call! Is it worth adding a test for the case where we didn't get as far as running the test-item?

NHDaly:
That's covered in the DontPass.jl tests already 👍 I was going to add a test for it, but there's already one there. Should I call that out explicitly?
```
        # This is the total number of testitems in DontPass.jl, minus those with
        # nonexistent testsetups, or errors in the testsetup.
        @test @atomic(end_expr_count.x) == 7
```

nickrobinson251:
ah great, thanks! no, that's good as is -- i just didn't read the test carefully enough 😊 

looks like some CI jobs need to be retried, so i'll do that. unfortunately our tests are a little flaky 😭 😭

PR #169, opened 2024-08-07
Move GC out of `test_end_expr` to its own keyword `gc_between_testitems`

nickrobinson251:
See https://github.com/JuliaTesting/ReTestItems.jl/pull/160#issuecomment-2273078012

PR#160 tidied up when we're invoking GC manually, and made it more user-controllable. This PR moves the user-control to be separate from the `test_end_expr`, in order to avoid users being forced to think about GC if they want to use `test_end_expr`. 

Marking this a bug-fix, since the last release contained PR#160 and means that anyone with an existing `test_end_expr` running with multiple workers would have experienced the complete removal of GC between test-items, whereas we only wanted them to benefit from the stream-lined invocation of GC (i.e. reduced to only being invoked if there are multiple workers, and only a single full pass).

### DIFF
--- a/src/ReTestItems.jl
+++ b/src/ReTestItems.jl
@@ -165,8 +165,6 @@ will be run.
 - `worker_init_expr::Expr`: an expression that will be run on each worker process before any tests are run.
   Can be used to load packages or set up the environment. Must be a `:block` expression.
 - `test_end_expr::Expr`: an expression that will be run after each testitem is run.
-  By default, the expression is `GC.gc(true)`, when `nworkers > 1`, to help with memory pressure,
-  otherwise it is a no-op.
   Can be used to verify that global state is unchanged after running a test. Must be a `:block` expression.
 - `memory_threshold::Real`: Sets the fraction of memory that can be in use before a worker processes are
   restarted to free memory. Defaults to $(DEFAULT_MEMORY_THRESHOLD[]). Only supported with `nworkers > 0`.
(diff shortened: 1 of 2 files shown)


src/ReTestItems.jl
```diff
@@ -388,11 +386,12 @@ function _runtests_in_current_env(
             # Now all workers are started, we can begin processing test items.
             @info "Starting running test items"
             starting = get_starting_testitems(testitems, nworkers)
+            gc_between_testitems = nworkers > 1
```
nickrobinson251, line 389: alternatively, we could expose a user-facing keyword to `runtests`, `gc_between_testitems`, which defaults to `nworkers>1`?

src/ReTestItems.jl
```diff
@@ -165,9 +165,13 @@ will be run.
 - `worker_init_expr::Expr`: an expression that will be run on each worker process before any tests are run.
   Can be used to load packages or set up the environment. Must be a `:block` expression.
 - `test_end_expr::Expr`: an expression that will be run after each testitem is run.
-  By default, the expression is `GC.gc(true)`, when `nworkers > 1`, to help with memory pressure,
-  otherwise it is a no-op.
   Can be used to verify that global state is unchanged after running a test. Must be a `:block` expression.
+- `gc_between_testitems::Bool`: If `true`, a full garbage collection (GC) will be run after each test item is run.
+  Defaults to `nworkers > 1`, i.e. `true` when running with multiple worker processes, since multiple worker processes
+  cannot coordinate to trigger Julia's GC, and should not be necessary to invoke the GC directly if running with without
```
nickrobinson251, line 171:
```suggestion
  cannot coordinate to trigger Julia's GC, and it should not be necessary to invoke the GC directly if running with without
```

Drvi (approved):

src/ReTestItems.jl
```diff
@@ -165,9 +165,13 @@ will be run.
 - `worker_init_expr::Expr`: an expression that will be run on each worker process before any tests are run.
   Can be used to load packages or set up the environment. Must be a `:block` expression.
 - `test_end_expr::Expr`: an expression that will be run after each testitem is run.
-  By default, the expression is `GC.gc(true)`, when `nworkers > 1`, to help with memory pressure,
-  otherwise it is a no-op.
   Can be used to verify that global state is unchanged after running a test. Must be a `:block` expression.
+- `gc_between_testitems::Bool`: If `true`, a full garbage collection (GC) will be run after each test item is run.
+  Defaults to `nworkers > 1`, i.e. `true` when running with multiple worker processes, since multiple worker processes
+  cannot coordinate to trigger Julia's GC, and it should not be necessary to invoke the GC directly if running with without
```
nickrobinson251, line 171:
```suggestion
  cannot coordinate to trigger Julia's GC, and it should not be necessary to invoke the GC directly if running without
```

PR #170, opened 2024-08-07
Filter testitems from AST before `eval`

nickrobinson251:


Drvi:
FWIW, [here](https://github.com/JuliaTesting/ReTestItems.jl/tree/td-ast-filter) is my very hacky implementation that skips eval , which does show a very significant perf improvement (but it does cheat by skipping over `@eval` and `@static` blocks and doesn't handle `@test_rel`s properly). Even with this limitation, I think it proves that successfully avoiding eval while filtering can help a lot (1.8secs to 0.3secs when I tried it again today).

nickrobinson251:
ooo very interesting -- i think the key is filtering the testitems out with `nothing` rather than an empty expression `:()`

```julia
julia> code1 = Union{Expr,Nothing}[:() for _ in 1:100_000];

julia> code2 = Union{Expr,Nothing}[nothing for _ in 1:100_000];

julia> @time for c in code1
           Core.eval(Main, c)
       end
  3.848924 seconds (3.60 M allocations: 164.779 MiB, 0.62% gc time)

julia> @time for c in code2
           Core.eval(Main, c)
       end
  0.015101 seconds (298.98 k allocations: 4.562 MiB)
```

i also like the `ShouldRun` clean-up on your branch !

Drvi:
I can't seem to replicate the 1 second speedup on RAICode:/

```julia
@btime runtests("test", name="aalkbjlhkbkhjvkhfcjgfca")
# PR:     1.533 s (6831030 allocations: 431.75 MiB)
# 1.24.0: 1.633 s (8192533 allocations: 497.06 MiB)
```
It seems that the remaining overhead is coming from `eval` of `@test_rel`s, this is the same issue I hit with my prototype. 
With these changes [diff](https://gist.github.com/Drvi/4b758542c3be1b8a572c9ec7ffb68eea), I was able to get quite nice performance:
```julia
# diff:  301.330 ms (3016961 allocations: 240.58 MiB)
```

What do you think? Is it worth special-casing test-rel like this? What would the alternative be?

src/filtering.jl
```diff
@@ -0,0 +1,133 @@
...
+# Any other macro calls (e.g. `@testsetup`) are left as is.
+# If the `@testitem` call is not as expected, it is left as is so that it throws an error.
+#
+# Replacing the expression with `nothing` effectively removes it from the file.
+# `Base.include` will still call `Core.eval(M, nothing)` which has a tiny overhead,
+# but less than `Core.eval(M, :())`. We could instead replace `Base.include` with a
+# custom function that doesn't call `Core.eval(M, expr)` if `expr === nothing`, but
+# using `Base.include` means we benefit from improvements made upstream rather than
+# having to maintain our own version of that code.
+function filter_testitem(f, expr)
```
Drvi, line 60: I got confused a little because there are two points at which we filter test items -- here at the AST level and then later in the `FilteredVector` level when we filter on the `eval`d `TestItem`. Let's maybe explain here, the things that need to be expanded, like:
```julia
@testitem "name with interpolation $(VERSION)" begin ... end
```
get filtered later and where? Because I think I'll forget how things work again:)

nickrobinson251: ooo, name with interpolation is interesting... i've never thought of us as (officially) supporting that.

i always thought we'd (one day) require files only have `@testitem` and `@testsetup` calls at the top-level, and that `name` and `tags` must be a static

But yes, you're quite right, for things that need to be expanded before filtering (`@test_rel`, `@eval ...`) we have to filter later in the `FilteredVector`, which is why (one day) i'd like us to not support those things at all and do all filter on the AST. Will add more comments explaining this.

nickrobinson251:
> Is it worth special-casing test-rel like this? What would the alternative be?

i've been reluctant to special-case `@test_rel` in ReTestItems.jl since this is an open-source test framework, whereas `@test_rel` is an ugly RAI-internal hack... 

BUT the compromise instead has to be allow any top-level macro-call (so as to allow `@test_rel`) rather than restricting to just `@testitem` and `@testsetup`... and that has led to other unintended uses, like this complete mis-use of ReTestItems:
```julia
@eval begin                                                                                        
    num_groups = 32
    for group_id in 1:num_groups
        eval(quote
            @testitem "MapOperator all natives - group $($group_id)" begin
            end
        )
    end
end
```

So, currently in this PR i stick with the (bad) status quo where we support arbitrary macrocalls at the top-level (assuming the create `TestItem`s) and so we have to still have the `FilteredVector`, and we pay the latency cost incurred by having to eval these macrocalls (even though the resulting TestItems might all be filtered out).

But we could instead disallow all macro-calls except `@testitem` and `@testsetup` and `@test_rel` (i.e. add a RAI-specific hack here), which as you say would allow us to write something that works on the `@test_rel`-AST... 
And we'd have to go edit the RAI codebase to rid it of the evil `@eval` code too.

idk what's best.

Drvi (approved):
I think this is great! A bunch of folks should get a nice latency reduction out of this:)

For RAI though, I think the transition to this new version of ReTestItems will be slightly challenging since we now disallow top-level `@eval` and I think the macro nesting with `@test_rel` cause some issues for the filtering logic, e.g. now I get these errors from `@test_rel`:
```
`tags` keyword must be passed a collection of `Symbol`s, got: $(Expr(:escape, :([:ring2, :integration])))
```
(I've tweaked the error message a bit to see what we got for `tags`).

Lastly, I think for RAI specifically, a good chunk of the latency reduction is still on the table, since currently, we eval all of the `@test_rel`s. It would be great if we could also try to extract the name and tags from these (many of them do have names and tags so we we should be able to skip eval often), but that can wait for a later PR.

nickrobinson251:
> we now disallow top-level @eval

i'll make a PR to update the codebase to remove `@eval`. I think the upfront pain is worth it to stop this mis-use before it spreads further 😛 

> It would be great if we could also try to extract the name and tags from these (many of them do have names and tags so we we should be able to skip eval often), but that can wait for a later PR.

agreed 👍 -- just want to land that in a follow-up PR as you say 😁 

> I think the macro nesting with @test_rel cause some issues for the filtering logic

this bit i didn't quite follow -- is it separate from the usage of `@eval`? How can i repro this?

nickrobinson251:
> I think the macro nesting with @test_rel cause some issues for the filtering logic

found what was wrong -- thanks for the pointer!

PR #174, opened 2024-08-15
Replace `relpath` with cheaper function

nickrobinson251:
Since we call `relpath` for every directory and file in the directory tree before any tests start, we want this to be as cheap as possible (to minimise latency). Hopefully this new function is valid on Windows, but I've no machine to test that :/

(micro-optimisation only noticable after #170)

### DIFF
--- a/src/ReTestItems.jl
+++ b/src/ReTestItems.jl
@@ -633,6 +633,20 @@ function is_testsetup_file(filepath)
     )
 end
 
+# like `relpath` but assumes `path` is nested under `startdir`, else just returns `path`
+function nestedrelpath(path, startdir)
+    path == startdir && return "."
+    relp = chopprefix(path, startdir)
+    sep = Base.Filesystem.path_separator
+    if endswith(startdir, sep)
+        return relp
+    elseif startswith(relp, sep)
+        return chopprefix(relp, sep)
+    else
+        return path
+    end
+end
+
 # is `dir` the root of a subproject inside the current project?
 function _is_subproject(dir, current_projectfile)
     projectfile = _project_file(dir)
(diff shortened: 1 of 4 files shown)


src/ReTestItems.jl
```diff
@@ -640,6 +640,20 @@ function is_testsetup_file(filepath)
...
+    relp = chopprefix(path, startdir)
+    sep = Base.Filesystem.path_separator
+    if endswith(startdir, sep)
+        return relp
+    elseif startswith(relp, sep)
+        return chopprefix(relp, sep)
+    else
+        return path
+    end
+end
```
Drvi, line 655:
```suggestion
function nestedrelpath(path, startdir)
    path == startdir && return SubString(".", 1:1)
    relp = chopprefix(path, startdir)
    sep = Base.Filesystem.path_separator
    if endswith(startdir, sep)
        return relp
    elseif startswith(relp, sep)
        return chopprefix(relp, sep)
    else
        return @view(path[1:end])
    end
end
```

With this suggestion, the function becomes type stable for `String` inputs (always returns a `SubString{String}`) and I get 0 allocs:

```julia
julia> @btime nestedrelpath($path, $("test"))
  54.048 ns (0 allocations: 0 bytes)
"dir/foo_test.jl"
```

nickrobinson251: thanks <3

Drvi (approved):
LGTM! I am not sure what the speedup is, but it seems like a good idea, thanks :+1:

src/ReTestItems.jl
```diff
@@ -693,7 +709,7 @@ function include_testfiles!(project_name, projectfile, paths, ti_filter::TestIte
             if !(is_testsetup_file(filepath) || (is_test_file(filepath) && is_requested(filepath, paths)))
                 continue
             end
-            fpath = relpath(filepath, project_root)
+            fpath = nestedrelpath(filepath, project_root)
             file_node = FileNode(fpath, ti_filter; report, verbose=verbose_results)
```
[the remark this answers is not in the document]
nickrobinson251: i think i didn't do this because i thought some package was using `FileNode` directly (despite it being internal)

but now i can't find such package so maybe would be okay to change

PR #179, opened 2024-09-09
Do all filtering of testitems at AST level

nickrobinson251:
Follow-up to #170 that allows us to remove the `FilteredVector` and do all filtering of testitems before `eval`'ing any code, this is possible (after #170) if we special-case the codepath we have to support RAI usage. This adds a bit more RAI-specific code to this package, but (i) speeds up RAI's use-case a lot and (ii) tidies up the filtering story in this package (since we can drop the codepaths that were doing filtering after `eval`ing code)

On the RAI test codebase this speeds up filtering of test-items considerably. Here's how long it takes to filter out all the tests (e.g. with `runtests(; name="doesnotexist", tags=[:nosuchtag])`)
- on v1.25.1:  ~10 seconds 
- after PR#170: ~8 seconds
- with this PR: ~1.3 seconds

### DIFF
--- a/src/filtering.jl
+++ b/src/filtering.jl
@@ -38,17 +38,6 @@ function (f::TestItemFilter)(expr::Expr)
     return expr
 end
 
-function _throw_not_macrocall(expr)
-    # `Base.include` sets the `:SOURCE_PATH` before the `mapexpr` is first called
-    file = get(task_local_storage(), :SOURCE_PATH, "unknown")
-    msg = """
-    Test files must only include `@testitem` and `@testsetup` calls.
-    In $(repr(file)) got:
-        $(Base.remove_linenums!(expr))
-    """
-    error(msg)
-end
-
 # Filter out `@testitem` calls based on the `name` and `tags` keyword passed to `runtests`.
 # Any other macro calls (e.g. `@testsetup`) are left as is.
 # If the `@testitem` call is not as expected, it is left as is so that it throws an error.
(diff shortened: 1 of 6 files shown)


src/filtering.jl
```diff
@@ -104,6 +112,32 @@ end
 # Macro used by RAI (corporate sponsor of this package)
 # TODO: drop support for this when RAI codebase is fully switched to ReTestItems.jl
 const ___RAI_MACRO_NAME_DONT_USE = Symbol("@test_rel")
+function __filter_rai(f, expr)
```
nickrobinson251, line 115: FYI this duplicates a lot of code from `try_get_tags` but i just really want this code to be trivial to delete (when the day comes) so i'd rather repeat some very similar logic than tie it into a function that we're probably going to keep

Drvi (approved):
People are going to notice this one:) Very nice, Nick!!

I tried it on RAICode and it was much faster! I did run into this one issue when I tried to run a single top-level `@test_rel`:
```julia
julia> runtests("test"; name="constant-folding-old-periods-old-1")
  Activating project at `~/_proj/raicode_master3`
[ Info: Scanning for test items in project `RAICode` at paths: test
┌ Warning: Encountered duplicate @testsetup with name: `LoadJSONSetup`. Replacing...
└ @ ReTestItems ~/.julia/packages/ReTestItems/NieGI/src/ReTestItems.jl:663
ERROR: TaskFailedException

    nested task error: LoadError: AssertionError: `tags` keyword must be passed a collection of `Symbol`s, got: $(Expr(:escape, :([:ring1, :integration]))) (Expr)
    Stacktrace:
      [1] var"@testitem"(__source__::LineNumberNode, __module__::Module, nm::Any, exs::Vararg{Any})
        @ ReTestItems ~/.julia/packages/ReTestItems/NieGI/src/macros.jl:264
...
```
(I has to comment out the top-level evals first)

nickrobinson251:
Thanks for pushing us to do this!

> one issue when I tried to run a single top-level @test_rel

yeah, this is a bug in RAI's `@test_rel` that i'll fix when updating to this new ReTestItems.jl release

PR #193, opened 2024-10-17
Run CI tests on Julia v1.11 too

nickrobinson251:
Drops v1.9 to avoid using too many resources. Keep v1.8 since that's the oldest we aim to support (and if that passes we can be fairly confident v1.9 works too).

Couple tests had to be updated:
- reference tests needed making less sensitive to how UndefVarError are printed
- behaviour of testing for `UndefVarError` broke in v1.11 https://github.com/JuliaLang/julia/issues/54082

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -18,8 +18,8 @@ jobs:
       matrix:
         version:
           - '1.8'
-          - '1.9'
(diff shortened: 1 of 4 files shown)


Drvi (approved):

NHDaly (approved):
coool

PR #195, opened 2024-10-22
Replace LoggingExtras.jl dep with our own debug macro

### LINKED ISSUE #135: Drop LoggingExtras.jl dependency
ReTestItems.jl having dependencies means we require users' code be compatible with those dependencies, specifically the version of those dependencies which ReTestItems.jl is using

Currently we only have 1 (non-stdlib) dependency, which is the LoggingExtras.jl v1... what do we use it for? is it just `@debugv`? if so, it'd be easy for us to roll our own version of that and drop the dependency

nickrobinson251:
close #135 (also fix this breakage https://github.com/JuliaTesting/ReTestItems.jl/pull/197#issuecomment-2579753553)


src/debug.jl
```diff
@@ -0,0 +1,37 @@
...
+
+"""
+    @debugv 1 "msg"
+"""
+macro debugv(level::Int, messsage)
+    mod = pkgdir(@__MODULE__)
+    quote
+        if DEBUG_LEVEL >= $level
+            _file = $String($(QuoteNode(__source__.file)))
+            _file = $relpath(_file, $mod)
```
quinnj, line 31: 2 file keyword args here?

NHDaly (approved):
LGTM

NHDaly:
And welcome back from your time off! :)

PR #197, opened 2025-01-08
fix windows failure

willow-ahrens:
fixes a windows failure due to asserting Int64 when the architecture may be Int32

```
 ERROR: LoadError: TaskFailedException

    nested task error: LoadError: MethodError: Cannot `convert` an object of type 
      Base.RefValue{Int32} to an object of type 
      Base.RefValue{Int64}
    
    Closest candidates are:
      convert(::Type{T}, ::T) where T
       @ Base Base.jl:84
      Base.RefValue{T}(::Any) where T
       @ Base refvalue.jl:8
    
    Stacktrace:
      [1] TestItem(number::Base.RefValue{Int32}, name::String, id::String, tags::Vector{Symbol}, default_imports::Bool, setups::Vector{Any}, retries::Int32, timeout::Nothing, skip::Bool, failfast::Nothing, file::String, line::Int32, project_root::String, code::Expr, testsetups::Vector{TestSetup}, workerid::Base.RefValue{Int32}, testsets::Vector{Test.DefaultTestSet}, eval_number::Base.RefValue{Int32}, stats::Vector{ReTestItems.PerfStats}, scheduled_for_evaluation::ReTestItems.ScheduledForEvaluation)
```

### DIFF
--- a/src/macros.jl
+++ b/src/macros.jl
@@ -112,7 +112,7 @@ those test in their own module.
 Should only be created via the `@testitem` macro.
 """
 struct TestItem
-    number::Base.RefValue{Int64} # populated by runtests coordinator once all test items are known
+    number::Base.RefValue{Int} # populated by runtests coordinator once all test items are known
     name::String
     id::String # in case file/name isn't a sufficiently stable identifier for reporting purposes
     tags::Vector{Symbol}


nickrobinson251 (approved):
thanks!

willow-ahrens:
I'd like to add windows CI to this PR to see if it works there, is that alright?

nickrobinson251:
> is that alright?

Yes, but let's only add it for a single Julia version (so we only add a single extra job).

willow-ahrens:
I'm not sure why all the other tests would fail here, but the tests on windows may indicate that some additional work is needed to support windows.

nickrobinson251:
> I'm not sure why all the other tests would fail here

Look like tests were broken by https://github.com/JuliaLogging/LoggingExtras.jl/pull/94

I guess that means we should finally do https://github.com/JuliaTesting/ReTestItems.jl/issues/135 / https://github.com/JuliaTesting/ReTestItems.jl/pull/195

PR #204, opened 2025-03-17
Update github actions versions

nickrobinson251:


.github/workflows/CI.yml
```diff
@@ -26,12 +26,12 @@ jobs:
...
+      - uses: actions/checkout@v4
+      - uses: julia-actions/setup-julia@v2
         with:
           version: ${{ matrix.version }}
           arch: ${{ matrix.arch }}
-      - uses: actions/cache@v2
+      - uses: actions/cache@v4
         env:
           cache-name: cache-artifacts
         with:
```
quinnj, line 37: There's a nice julia-specific cache action, so you can just replace this whole action block with `- uses: julia-actions/cache@v2`

nickrobinson251:
going with this as-is to get CI back online. Will make a separate PR to use `julia-actions/cache@v2`

PR #212, opened 2025-05-23
Hacky prototype for running test failures first

nickrobinson251:


src/ReTestItems.jl
```diff
@@ -407,6 +424,9 @@ function _runtests_in_current_env(
             ctx = TestContext(proj_name, ntestitems)
             # we use a single TestSetupModules
             ctx.setups_evaled = TestSetupModules()
+            if cfg.failures_first && !isempty(GLOBAL_TEST_STATUS)
+                sort!(testitems.testitems; by=_status_when_last_seen)
+            end
```
Drvi, line 429: Since you mention the additional overhead of sorting, how bad is it? Perhaps we could exploit the fact that we have three discrete values to sort by and try an algorithm that traverses the array once, swapping failed elements to the front and passed elements to the back? Or we could use 3 arrays instead of one and then do something like `Iterators.flatten` to iterate the items in the correct order.

src/ReTestItems.jl
```diff
@@ -245,6 +255,10 @@ will be run.
   Defaults to the value passed to the `failfast` keyword.
   If a `@testitem` sets its own `failfast` keyword, then that takes precedence.
   Note that the `testitem_failfast` keyword only takes effect in Julia v1.9+ and is ignored in earlier Julia versions.
+- `failures_first::Bool=false`: if `true`, first runs test items that failed the last time
+  they ran, followed by new test items, followed by test items that passed the
+  last time they ran. This is currently only supported when `nworkers=0`.
+  Can also be set using the `RETESTITEMS_FAILURES_FIRST` environment variable.
```
Drvi, line 261: Why isn't this supported with multiple workers? It seems to me the coordinator controls the order of the queue and knows the result of the test item.

Drvi (commented):
Thanks, Nick! I left a couple of comments / questions.

I think ideally, we'd generate an artifact that one could download from CI, and feed it to `ReTestItems` to reproduce the failures. To me this is the most likely scenario this feature would be useful.

src/ReTestItems.jl
```diff
@@ -23,6 +23,16 @@ else
     const errmon = identity
 end
 
+# Used by failures_first to sort failures before unseen before passes.
+const GLOBAL_TEST_STATUS = Dict{String,UInt8}()
```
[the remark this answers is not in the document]
nickrobinson251: the `id` is a string already before this PR, because... we needed it to be for the `@test_rel` macro for some reason i can't remember, and i didn't want to change it in this PR 

We could change `id` to be ` UInt64` and update this here accordingly, but i'd rather do that as a separate PR

src/ReTestItems.jl
nickrobinson251: i'd say for 10,000 test items it's still acceptable tbh
```julia
julia> using Random, BenchmarkTools

julia> function bench_sort(total)
           cached = total ÷ 2
           order = Dict(string(rand(UInt64)) => rand(UInt8[0,1,2]) for _ in 1:cached);
           sort_func(k) = get(order, k, UInt8(0));
           rest = [string(rand(UInt64)) for _ in 1:(total-cached)]
           items = [keys(order)..., rest...][randperm(total)]
           @btime sort!($items; by=$sort_func)
       end
bench_sort (generic function with 1 method)

julia> foreach([100, 1_000, 10_000, 100_000]) do x
           bench_sort(x)
       end
  1.804 μs (0 allocations: 0 bytes)
  27.500 μs (0 allocations: 0 bytes)
  552.916 μs (0 allocations: 0 bytes)
  5.281 ms (0 allocations: 0 bytes)

julia> VERSION
v"1.10.10"
```

src/ReTestItems.jl
nickrobinson251: that's a very different implementation, because it flips where state is stored. right now each testitem knows what number it is and if it's available (or already taken), and we simply ask for the next available one, and there's nowhere that know the state of each worker. whereas in this proposal there would be a central manager. So that's a big change, and i don't think this feature requires it.

PR #231, opened 2025-11-01
Default to `failures_first=true`

nickrobinson251:
We've had feedback from users that this option is really helpful, so let's make it the default.

Here's a data point for the overhead it adds for 1,000 testitems:
```julia
julia> for fileno in 1:100
           fname = "tmp/file$(fileno)_test.jl"
           open(fname, "w") do io
               for i in 1:100
                   tnum = (fileno-1)*100 + i
                   write(io, """
                       @testitem "$tnum" begin
                           @test $(rand(Bool))
                       end
                       """
                   )
               end
           end
       end

julia> runtests("tmp/") # run once to populate test statuses

julia> runtests("tmp/") # run again with the additional cost of sorting testitems
...
DEBUG @ ReTestItems.jl:429 | Sorted test items with failures first in 0.00694 seconds.
...
```
where i added this timing info like:
```diff
--- a/src/ReTestItems.jl
+++ b/src/ReTestItems.jl
@@ -420,10 +420,13 @@ function _runtests_in_current_env(
         (nworkers == 0 ? "" : " with $nworkers worker processes and $nworker_threads threads per worker.")
     try
         if cfg.failures_first && !isempty(GLOBAL_TEST_STATUS)
+            t0 = time()
             sort!(testitems.testitems; by=_status_when_last_seen)
             foreach(enumerate(testitems.testitems)) do (i, ti)
                 ti.number[] = i # reset number to match new order
             end
+            tt = time() - t0
+            @debugv 2 "Sorted test items with failures first in $(round(tt, digits=5)) seconds."
             is_sorted_queue = true
```

I added `ReTestItems.reset_test_status!()` to tests as needed. This isn't public API but now `failures_first=true` is the default i think more likely we will get a feature request for it -- let's see.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "ReTestItems"
 uuid = "817f1d60-ba6b-4fd5-9520-3cf149f6a823"
-version = "1.34.1"
+version = "1.35.0"
 
 [deps]
 Dates = "ade2ca70-3891-5945-98fb-dc099432e06a"
--- a/README.md
+++ b/README.md
@@ -133,6 +133,7 @@ If you want individual test-items to stop on their first test failure, but not s
 #### Running previous failures first
 
 By default `runtests` uses `failures_first=true` to run first any test-items that failed the last time they were run.
+This became the default in v1.35.0.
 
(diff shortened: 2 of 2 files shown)


NHDaly (approved):
❤️ This is turning into such a great little testing framework ❤️

PR #242, opened 2026-08-05
Add a `cost` keyword to `@testitem` and schedule by it

### LINKED ISSUE #241: Schedule long-running test items first via a `cost` keyword on `@testitem`
When running with `nworkers`, a long test item that gets picked up near the end of the run leaves every other worker idle while it finishes. The classic fix (LPT scheduling) is to start the longest jobs first — but ReTestItems has no way to know ahead of time which items are long. Proposal: let a test item declare it, and schedule by it:

```julia
@testitem "slow integration test" cost=450 begin
    @test long_running_thing()
end
```

- `cost` is a number of nominal seconds. Only relative sizes matter, so a rough, consistent scale within a project is enough.
- Items that declare a cost are claimed before those that don't, most expensive first; items without a cost keep their current (file) order after them.
- A cost can also be a function of the run configuration, e.g. `cost=(cfg -> 15 + 450 / cfg.nworker_threads)` for work the item parallelizes across its worker's threads. It's called once per item in the coordinator, before anything runs, and replaced by the number it returns.

Related: #217, which sketches strict priorities (higher priority *completes* before lower starts, for tests that aren't independent). This is deliberately narrower: only a hint about task order. It composes naturally with #172/`failures_first`: failures still sort first (feedback latency beats makespan), then by cost within each group. For prior art in the "declared cost" style: Bazel's `size` attribute on test targets; the pytest ecosystem does the measured-timings variant (e.g. pytest-split).

Design points I'd like input on:

- Naming: `cost`? (`duration`, `weight`, `size`?)
- When any cost is declared the queue is sorted, so workers start at the front of the queue rather than at evenly spaced positions (same regime `failures_first` already uses on re-runs). That trades a bit of per-worker file locality — shared `@testsetup`s may get evaluated on more workers — for makespan; unlike `failures_first` it applies to every run, since costs live in the source.
- Undeclared costs sort last rather than being assigned a default and interleaved — predictable, but it means the feature only helps if your slowest items are the annotated ones.

A side benefit of the proposed implementation: the sort key gets a unique tie-break, which makes the `failures_first` ordering deterministic (it currently depends on an unstable sort).

jey:
Close #241

Adds a `cost` keyword to `@testitem` declaring roughly how long the item takes to run, in nominal seconds, and schedules by it: items that declare a cost are claimed before those that don't, most expensive first, so a long item can't start near the end of the run and leave the other workers idle waiting for it.

```julia
@testitem "slow integration test" cost=450 begin
    @test long_running_thing()
end
```

Details:

- The queue is sorted by `(status_when_last_seen, -cost, number)`; `number` is unique, so the order is deterministic regardless of sort stability. This also makes the existing `failures_first` ordering deterministic, which it previously was not. `failures_first` takes precedence over cost within the key.
- Items with no declared cost sort after all items that do, in the order they would otherwise have run in. A run with no costs and no cached failure status takes the exact same path as before (unsorted queue, evenly spaced worker starts).
- When the queue is sorted, workers start at the front rather than at evenly spaced positions, so the most expensive items start immediately.
- A cost can also be a function of the run configuration, for items whose duration depends on it, e.g. `cost=(cfg -> 15 + 450 / cfg.nworker_threads)`. It receives `(; nworkers::Int, nworker_threads::Int)` and is called exactly once per item, in the coordinator, before any item runs; the returned number replaces the function, so no user function is ever serialized to a worker. `nworker_threads` is the thread count items will actually run with (the setting, or `Threads.nthreads()` when serial).
- Costs are validated to be finite and non-negative (`Bool` rejected); a function must return a `Real` or `nothing`; errors while evaluating a cost name the test item.

Tests cover the macro parsing/validation, the sort (including ties, precedence with `failures_first`, and worker starting positions), cost-function resolution, and end-to-end run order via `runtests` with `nworkers` 0/1/2. README and docstring updated; version bumped to 1.36.0.

Questions:

- Happy to rename the keyword if `cost` isn't the right word.
- Should there be a run-level switch on `runtests` to ignore declared costs (keyword + env var, like `failures_first`)? As is, declaring one cost opts the whole suite into the sorted-queue regime on every run.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "ReTestItems"
 uuid = "817f1d60-ba6b-4fd5-9520-3cf149f6a823"
-version = "1.35.2"
+version = "1.36.0"
 
 [deps]
 Dates = "ade2ca70-3891-5945-98fb-dc099432e06a"
--- a/README.md
+++ b/README.md
@@ -229,6 +229,30 @@ end
 
 If a test-items set the `failfast` then that value takes precedence over the `testitem_failfast` keyword passed to `runtests`.
 
+#### Running expensive test-items first
+
+If a test-item takes much longer to run than the others, it can say so using the `cost` keyword.
+Test-items that declare a cost are run before those that don't, most expensive first.
+
+```julia
+@testitem "slow integration test" cost=450 begin
+    @test long_running_thing()
+end
+```
+
+Starting the most expensive test-items first shortens the whole test run, since a long test-item that starts near the end of the run leaves every other worker idle waiting for it to finish.
+A cost is a number of nominal seconds. Only the relative size of costs matters, so costs need not be accurate — a rough measurement is enough to get most of the benefit — but they should be on a consistent scale within a project.
+
+A cost can also be computed from the configuration of the test run, by passing a function.
+It is called once per test-item, before any test-item runs, with a `NamedTuple` holding `nworkers::Int`, the number of worker processes (`0` when tests run serially), and `nworker_threads::Int`, the number of threads each test-item will run with.
+For example, a test-item with 15 seconds of fixed cost plus 450 seconds of work that its worker can parallelize across threads:
+
+```julia
+@testitem "slow integration test" cost=(cfg -> 15 + 450 / cfg.nworker_threads) begin
+    @test long_running_thing()
+end
+```
+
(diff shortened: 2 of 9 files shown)


jey:
Marking this ready for review since the two CI test failures appear to be spurious.

